PR #13, opened 2018-07-17
[build] Add Dune support.

ejgallego:
Just in case it could be useful.

Note that the `opam` file will not be there for the next version of
Dune, but generated automatically.


Zimmi48:
Many thanks! I'll test this setup ASAP.

PR #28, opened 2018-12-21
Do merge in fresh worktree

SkySkimmer:
This should ensure fresh state better.

I split the merge command to a new sh file as doing concatenations in
ocaml is a pain and stops me from using shell linters etc.

This probably doesn't work as the call to make_ancestor.sh is using
PATH and I don't think it's in PATH, how to fix?

### DIFF
--- a/bot.ml
+++ b/bot.ml
@@ -65,11 +65,7 @@ let git_delete repo remote_branch_name =
   git_push ~force:false repo "" remote_branch_name
 
 let git_make_ancestor ~base ref2 =
-  Printf.sprintf "git checkout -f %s" ref2
-  |&& Printf.sprintf
-        "git merge %s -m \"Bot merge $(git rev-parse %s) into $(git rev-parse \
-         %s)\" || { git merge --abort; false; }"
-        base base ref2
+  Printf.sprintf "../make_ancestor.sh %s %s" base ref2
 
 let extract_commit json =
   let open Yojson.Basic.Util in
--- a/make_ancestor.sh
+++ b/make_ancestor.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+# usage: make_ancestor.sh base ref
+# PWD must be in the relevant git repo
+# merges base in ref if necessary
+
+set -ex
+
+if [ $# != 2 ]; then >&2 echo Bad argument count; fi
+
+base=$1
+ref=$2
+
+basecommit=$(git rev-parse "$base")
+refcommit=$(git rev-parse "$ref")
+
+wtree=$(mktemp -d)
+
+( git checkout --detach HEAD # if ref is already checked out add worktree errors
+  git worktree add "$wtree" "$ref"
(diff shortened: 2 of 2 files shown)


Zimmi48:
Thanks, I was about to open an issue "Do merge in worktree" when I saw your PR :grinning:

bot.ml
```diff
@@ -65,13 +65,7 @@ let git_delete repo remote_branch_name =
...
 
 let git_make_ancestor ~base ref2 =
-  (* Ensure state is clean *)
-  "git rm .gitattributes && git reset --hard"
-  |&& Printf.sprintf "git checkout %s" ref2
-  |&& Printf.sprintf
-        "git merge %s -m \"Bot merge $(git rev-parse %s) into $(git rev-parse \
-         %s)\" || { git merge --abort; false; }"
-        base base ref2
+  Printf.sprintf "make_ancestor.sh %s %s" base ref2
```
Zimmi48: Just doing `./make_ancestor.sh` should work because all the commands are executed in the current working directory by default (this is how it's done in `backport_pr` below).

make_ancestor.sh
```diff
@@ -0,0 +1,29 @@
...
+
+base=$1
+ref=$2
+
+basecommit=$(git rev-parse "$base")
+refcommit=$(git rev-parse "$ref")
+
+wtree=$(mktemp -d)
+
+( git checkout --detach HEAD # if ref is already checked out add worktree errors
```
Zimmi48: It doesn't as long as `ref` is a commit hash. On the other hand, such a `git checkout` could be a source of problems so what about removing it.

bot.ml
SkySkimmer: >all the commands are executed in the current working directory by default

How does that work, isn't the working directory the git repo for coq?

make_ancestor.sh
SkySkimmer: We really want to check out the branch, not the commit, because we want to add a merge commit to it.

SkySkimmer: What problems would it cause?

bot.ml
Zimmi48: We clone coq in `./repo`. Thus most commands start with `cd_repo |&&`, including the one in which this function is called. So indeed, it should rather be:

```suggestion
  Printf.sprintf "../make_ancestor.sh %s %s" base ref2
```

One nice thing about your change is that it will allow once again considering to make the git repository bare (and only its worktrees are not). In this case, we can even put the git repository in the same location as the scripts instead of a sub-directory.

make_ancestor.sh
Zimmi48: >We really want to check out the branch, not the commit, because we want to add a merge commit to it.

Ah right.

>What problems would it cause?

None, I hope. Anyways, this could be removed in a next step when making the repository bare.

bot.ml
SkySkimmer: Changed.

PR #56, opened 2020-06-09
Fix project name for job actions.

### LINKED ISSUE #55: Job failures are not reported when project name differs from project slug on GitLab.
With @jtcoolen, we noticed that a project on GitLab can have a name which doesn't reflect correctly the project URL. In this case, the status checks for failed GitLab CI jobs are not reported to the right URL on GitHub.

jtcoolen:
This fixes #55 by extracting the project name from the repository url first, and falling back to the `project_name` key if it fails.

### DIFF
--- a/bot.ml
+++ b/bot.ml
@@ -472,8 +472,10 @@ let job_action json =
   let branch = json |> member "ref" |> to_string in
   let context = f "GitLab CI job %s (%s)" build_name (branch_or_pr branch) in
   let repo_full_name =
-    json |> member "project_name" |> to_string
-    |> Str.global_replace (Str.regexp " ") ""
+    let repo_url = json |> member "repository" |> member "url" |> to_string in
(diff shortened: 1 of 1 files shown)


bot.ml
```diff
@@ -472,8 +472,11 @@ let job_action json =
   let branch = json |> member "ref" |> to_string in
   let context = f "GitLab CI job %s (%s)" build_name (branch_or_pr branch) in
   let repo_full_name =
-    json |> member "project_name" |> to_string
-    |> Str.global_replace (Str.regexp " ") ""
+    let repo_url = json |> member "repository" |> member "url" |> to_string in
+    if not (string_match ~regexp:".*:\\(.*\\).git" repo_url) then
+    Stdio.printf "Could not match project name on repository url"
+    else ();
```
Zimmi48, line 478: In OCaml, you can skip `else ()` clauses.

jtcoolen: Indeed!

Issue #46, opened 2020-01-02
Authenticating to GitHub's API using username + password is deprecated.


Zimmi48:
All indirect uses of the following line to communicate with the GitHub API should be removed:

https://github.com/coq/bot/blob/e527d0554c8cadf122056ec0cbbf0dc874bb7ad6/bot.ml#L13

The GitHub access token should be used instead of the password.

More info at https://developer.github.com/v3/auth/#basic-authentication

Zimmi48:
This must be fixed by July 1st, 2020.

Zimmi48:
The new timeline is that this must be fixed by end of September: https://developer.github.com/changes/2020-02-14-deprecating-password-auth/

PR #57, opened 2020-06-22
Authenticate with access token.

jtcoolen:
This PR removes deprecated password/login authentication in favor of authentication tokens, fixing #46.

### DIFF
--- a/bot.ml
+++ b/bot.ml
@@ -6,17 +6,13 @@ open Utils
 
 let port = try Sys.getenv "PORT" |> int_of_string with Not_found -> 8000
 
-let username = Sys.getenv "USERNAME"
-
-let password = Sys.getenv "PASSWORD"
-
-let credentials = `Basic (username, password)
-
 let gitlab_access_token = Sys.getenv "GITLAB_ACCESS_TOKEN"
 
-let gitlab_header = [("PRIVATE-TOKEN", gitlab_access_token)]
+let gitlab_header = [("Private-Token", gitlab_access_token)]
 
 let github_access_token = Sys.getenv "GITHUB_ACCESS_TOKEN"
+
+let github_header = [("Authorization", "token " ^ github_access_token)]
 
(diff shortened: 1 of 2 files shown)


bot.ml
```diff
@@ -112,7 +108,7 @@ let headers header_list =
   Header.init ()
   |> (fun headers -> Header.add_list headers header_list)
   |> (fun headers -> Header.add headers "User-Agent" "coqbot")
-  |> fun headers -> Header.add_authorization headers credentials
+  |> (fun headers -> Header.add_list headers github_header)
```
Zimmi48, line 111: Instead of adding the github header here, you could do it every time `send_request` is used for GitHub (replace `[]` with `github_header`, just like for `gitlab_header`). That would be cleaner.

bot.ml
```diff
@@ -545,7 +541,7 @@ let job_action json =
     |> Lwt.async
   else if String.equal build_status "success" then (
     (fun () ->
-      get_status_check ~repo_full_name ~commit ~context
+      get_status_check ~repo_full_name ~commit ~context ~header_list:gitlab_header
```
jtcoolen, line 544: The `github_header` should be included in `header_list` too, right?

jtcoolen: No, the gitlab api is only used in this function call, so `gitlab_header` is not needed.

bot.ml
```diff
@@ -224,16 +219,17 @@ let handle_json action default body =
       Stdio.printf "Json type error: %s\n" err ;
       default
 
-let generic_get relative_uri ?(header_list = []) ~default json_handler =
+let generic_get relative_uri ~header_list ~default json_handler =
   let uri = "https://api.github.com/" ^ relative_uri |> Uri.of_string in
```
Zimmi48, line 223: Despite the `generic_get` name, this function is specific to the GitHub API, so it could already be configured with the `github_header` by default.

bot.ml
Zimmi48, line 544: Here you put `gitlab_header` but actually (as explained above), this is a call to the GitHub API.

PR #60, opened 2020-06-25
Moving the GitHub/GitLab queries and mutations to the relevant modules

jtcoolen:
After moving the GitHub/GitLab queries and mutations to the relevant modules, we then rewrite `add_rebase_label` to use GitHub's GraphQL API.


Zimmi48:
Sorry @jtcoolen, I've created again a conflict with your branch. But this one should be easy to fix.

jtcoolen:
Rebased the branch on top of master.

Issue #58, opened 2020-06-24
dune build --ignore-promoted-rules seems to not be working anymore


Zimmi48:
Running `dune build --ignore-promoted-rules` with no `bot-components/.github-token` was supposed to succeed and now it gives the following error message:

```
File "bot-components/dune", line 19, characters 0-187:
19 | (rule
20 |  (targets .graphqlconfig)
21 |  (deps
22 |   (:input .graphqlconfig.in)
23 |   .github-token)
24 |  (action
25 |   (with-stdout-to
26 |    %{targets}
27 |    (run sed "s/%%TOKEN%%/%{read:.github-token}/" %{input}))))
Error: No rule found for bot-components/.github-token
```

PR #64, opened 2020-07-01
Add (mode promote) to rule in dune file.

jtcoolen:
There was a missing `(mode promote)` to a rule in the dune file which caused an error during compilation, as reported in issue #58.


jtcoolen:
Your second proposition is more concise, one rule should suffice.

PR #65, opened 2020-07-07
Dockerfile to ease the build and deployment process

jtcoolen:
A Docker file to simplify the build and deployment process (ran the container on Heroku, it went through our integration tests without issues).
Our next step: automate the release and deployment of the server to Heroku with GitHub Actions.


Zimmi48:
cc @erikmd: it turns out that the issue when running in Docker had nothing to do with Docker but was that some (optional) opam dependencies were missing (compared to when building with Nix before deploying).

erikmd:
Great! :-)

Just a minor comment: have you tested replacing the line

https://github.com/coq/bot/blob/f9ce5a99ba5fa3a32370d2b64ac28848a11cfdda/Dockerfile#L44

(in shell form) with

```
CMD ["./bot.exe"]
```

(in exec form)?

this should be more efficient: no shell would be spawned and `bot.exe` would have PID 1, which could also have some positive impact on the stop time, cf. e.g. the PID 1 zombie reaping problem mentioned in the thread [Speed up docker-compose shutdown](https://stackoverflow.com/a/50536078/9164010)

jtcoolen:
That's an improvement indeed, thank you @erikmd. We took into account your piece of advice.

PR #74, opened 2020-07-16
Integrate with run-coq-bug-minimizer GitHub action.

jtcoolen:
This patch allows coqbot to reply to '@coqbot: minimize' messages with the results of @JasonGross' [Coq bug minimizer](https://github.com/JasonGross/run-coq-bug-minimizer).

Here is an example:
minimizer

Before merging this, we'd need to change the repository in the `coq_bug_minimizer` script to the original one (it uses a fork currently).

I also did some preliminary work to listen to check run events, so that coqbot could reply back once the check run starts, with a link to the running check.

### DIFF
--- a/bot.ml
+++ b/bot.ml
@@ -815,7 +815,11 @@ let callback _conn req body =
                          "git push https://%s:%s@github.com/%s.git --delete \
                           '%s'"
                          bot_name github_access_token repo_name branch_name)
-                  >>= fun _ -> Lwt.return () ))
+                  >>= function
+                  | Ok () ->
+                      Lwt.return ()
+                  | Error f ->
+                      Lwt_io.printf "Error: %s" f ))
             |> Lwt.async ;
             Server.respond_string ~status:`OK ~body:"" ()
         | _ ->
--- a/bot.ml
+++ b/bot.ml
@@ -14,6 +14,8 @@ let github_webhook_secret = Sys.getenv "GITHUB_WEBHOOK_SECRET"
 
 let bot_name = try Sys.getenv "BOT_NAME" with Not_found -> "coqbot"
 
+let bot_domain = f "%s.herokuapp.com" bot_name
+
 let bot_email =
   try Sys.getenv "BOT_EMAIL"
   with Not_found -> f "%s@users.noreply.github.com" bot_name
@@ -69,6 +71,22 @@ let git_make_ancestor ~base head =
       Error (f "git_make_ancestor script killed by signal %d." signal)
   | Unix.WSTOPPED signal ->
       Error (f "git_make_ancestor script stopped by signal %d." signal)
+
+let git_coq_bug_minimizer ~script ~comment_thread_id ~comment_author ~token
+    ~bot_name ~bot_domain =
+  f "./coq_bug_minimizer.sh '%s' %s %s %s %s %s" script comment_thread_id
+    comment_author token bot_name bot_domain
+  |> Lwt_unix.system
+  >|= fun status ->
+  match status with
+  | Unix.WEXITED 0 ->
+      Ok true (* push successful *)
+  | Unix.WEXITED code ->
+      Error (f "coq_bug_minimizer script exited with status %d." code)
+  | Unix.WSIGNALED signal ->
+      Error (f "coq_bug_minimizer script killed by signal %d." signal)
+  | Unix.WSTOPPED signal ->
+      Error (f "coq_bug_minimizer script stopped by signal %d." signal)
 
 let first_line_of_string s =
   if string_match ~regexp:"\\(.*\\)\n" s then Str.matched_group 1 s else s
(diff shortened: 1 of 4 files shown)


Zimmi48:
>I also did some preliminary work to listen to check run events, so that coqbot could reply back once the check run starts, with a link to the running check.

Great! But then, you'll need some webhook on the repo executing the minimizer?

JasonGross:
Alternatively, if you want, we can add another script to the run-minimizer that runs when `run.sh` starts, and creates the URL from GH Actions environment variables, and tells coqbot what to post

bot.ml
```diff
@@ -672,6 +690,34 @@ let callback _conn req body =
...
+              ~comment_author:comment_info.author ~bot_name ~bot_domain
+              ~token:github_access_token
+            >>= fun _ ->
+            GitHub_mutations.post_comment ~id:comment_thread_id
+              ~message:
+                (f
+                   "Hey @%s, the coq bug minimizer is running your script, \
+                    I'll come back to you with the results once it's done."
+                   comment_info.author)
+              ~token:github_access_token
```
Zimmi48, line 717: It would be best to not ignore the result of the previous action, and in particular, to not post this comment if the coq bug minimizer is not actually running anything because something went wrong. More generally, it would be best to use this `fun _ ->` pattern as little as possible. For instance, the first of the three here can simply be replaced by `fun () ->` I think.

bot.ml
```diff
@@ -750,6 +796,24 @@ let callback _conn req body =
...
+        let message = Str.matched_group 2 body in
+        match Str.split (Str.regexp " ") stamp with
+        | [id; author; repo_name; branch_name] ->
+            GitHub_mutations.post_comment ~token:github_access_token ~id
+              ~message:(f "@%s, %s" author message)
+            >>= fun () ->
+            execute_cmd
+              (f "git push https://%s:%s@github.com/%s.git --delete '%s'"
+                 bot_name github_access_token repo_name branch_name)
+            >>= fun _ -> Server.respond_string ~status:`OK ~body:"" ()
```
Zimmi48, line 813: Since the response doesn't depend on the results on the previous action, only on the fact that the parsing of the body went well, you can send it asynchronously. The git delete and post comment actions can also be run in parallel (cf. the `<&>` operator).

coq_bug_minimizer.sh
```diff
@@ -0,0 +1,34 @@
...
+branch_name="run-coq-bug-minimizer-$branch_id"
+
+wtree=$(mktemp -d)
+
+git fetch "https://github.com/$repo_name.git" "refs/heads/master:$branch_name"
+git worktree add "$wtree" "$branch_name"
+pushd "$wtree"
+
+printf "%s %s %s %s" "$comment_thread_id" "$comment_author" "$repo_name" "$branch_name" > coqbot-request-stamp
+printf "%s\n" "$script" > coqbot.sh
```
Zimmi48, line 26: Not necessarily something to change now, but we could consider creating files from OCaml instead, either as temporary files that get copied here, or by converting the whole script to OCaml as you proposed yourself.

coq_bug_minimizer.sh
```diff
@@ -0,0 +1,34 @@
...
+sed -i 's/\r$//g' coqbot.sh
+sed -i "s/COQBOT_URL='.*'/COQBOT_URL='https:\/\/$bot_domain\/coq-bug-minimizer'/" coqbot-config.sh
```
Zimmi48, line 28: If we don't plan to change the other parameters, what about removing them from `coqbot-config.sh` and writing the file in its entirety instead of relying on `sed`?

Zimmi48:
> Alternatively, if you want, we can add another script to the run-minimizer that runs when `run.sh` starts, and creates the URL from GH Actions environment variables, and tells coqbot what to post

Or, we could go even further and rely on these check run events for triggering the final comment as well instead of running a script from within the action. I don't have a strong opinion on this question.

coq_bug_minimizer.sh
jtcoolen: Yes, we'll do this as part of our next refactoring.

jtcoolen:
> > I also did some preliminary work to listen to check run events, so that coqbot could reply back once the check run starts, with a link to the running check.
> 
> Great! But then, you'll need some webhook on the repo executing the minimizer?

Yes, a webhook must be set up. Jason's above suggestion is fine too.

I've changed the repository in the `coq_bug_minimizer` script to the one from Jason, it should work provided coqbot has push rights on this repo.

coq_bug_minimizer.sh
JasonGross: I've replaced `COQBOT_URL=''` with `COQBOT_URL="$(cat "$DIR/coqbot.url")"` in this file permanantly.  So now you either should do
```suggestion
sed -i "s/COQBOT_URL=.*/COQBOT_URL='https:\/\/$bot_domain\/coq-bug-minimizer'/" coqbot-config.sh
```
or
```suggestion
echo "https://$bot_domain/coq-bug-minimizer" > coqbot.url
```

jtcoolen: Picked the second one.

JasonGross:
> @JasonGross What is your plan to document / advertise the feature?

That's a good question.  @achlipala might have good ideas.  My implicit (and not very good) plan was to just comment manually on any issues I noted that could benefit, and rely on word-of-mouth.  Ideas that come up:
1. Announce it on coqdev
2. Announce it on coq-club
3. Announce it on Zulip
4. Announce it on discourse
5. Add mention of it on the Coq website where error reporting goes
6. Add it to the issue template, in a way that coqbot will catch it if the user fills it out and such that coqbot will not catch it if the user leaves it empty in a comment or whatever
7. Add some heuristic to coqbot to notice when an issue (or comment) is made that is a likely candidate for minimization, maybe when the text contains a link, the word `make`, the name of a `.v` file, and does not contain an attachment nor any block of code consisting of anything other than a Coq error message.

I'm personally a fan of 6, now that I've thought of it, and I think 7 is probably too fragile, too much work, and too complicated.

Thoughts? Opinions?

jtcoolen:
It makes sense to advertise this feature in the issue template, here is a proposition: https://github.com/coq/coq/pull/12706.

PR #76, opened 2020-07-21
Update coq_bug_minimizer.sh

JasonGross:
I've transferred run-coq-bug-minimizer to coq-community


Zimmi48:
Please always open an issue to propose a transfer (even if you have my agreement) so that other community members are made aware of the transfer and its interest for the community... Not a big deal though.

PR #77, opened 2020-07-22
Configuration file

jtcoolen:
This patch brings a configuration file to parametrize @coqbot 

@coqbot users can also put a `coqbot.toml` file in their GitHub repositories to specify the location of the corresponding GitLab repository running the CI jobs:
```
[mapping]
gitlab = "owner/repo"
```

What remains to be done:
- read the URLs of the GitHub and GitLab instances from the configuration (we could even give the URL of the instances in the mappings)


Zimmi48:
Note: as we discussed, we should probably build a different Docker image for the continuous deployment of `master` to Heroku and for the one of new releases to GitHub Packages (the latter would not include `coqbot.toml` or would include a stripped down version).

src/dune
```diff
@@ -3,7 +3,8 @@
  (public_name bot)
  (package coq-bot)
  (promote (until-clean))
- (libraries base bot-components cohttp-lwt-unix stdio yojson))
+ (modules bot config)
```
Zimmi48, line 6: What happens if you don't list the modules explicitlly?

src/config.ml
```diff
@@ -0,0 +1,106 @@
...
+  match subkey_value toml_data "github" "webhook_secret" with
+  | None ->
+      Sys.getenv_exn "GITHUB_WEBHOOK_SECRET"
+  | Some secret ->
+      secret
+
+let bot_name toml_data =
+  Option.value_map
+    (subkey_value toml_data "bot" "name")
+    ~f:String.of_string ~default:"coqbot"
```
Zimmi48, line 52: For now, one can configure the bot name and email with environment variable. I'm OK with removing this possibility but let's not forget that the documentation should be updated to reflect that.

src/config.ml
```diff
@@ -0,0 +1,106 @@
...
+    | _ ->
+        None
+  in
+  Option.value_exn toml_mappings
+    ~message:"No mappings field found in toml config file."
+  |> parse_mappings
+
+let github_of_gitlab gl mapping = Hashtbl.find mapping gl
+
+let gitlab_of_github gh mapping = Hashtbl.find mapping gh
```
Zimmi48, line 106: I'm not quite sure what we gain by re-exporting these functions, except if you plan on having an abstract interface file where the types for GitHub to GitLab and GitLab to GitHub mappings are not the same.

jtcoolen:
We can leave the configuration of the GitHub/GitLab URLs for later, to concentrate on our refactoring.

README.md
```diff
@@ -223,6 +223,26 @@ these are configured in your Heroku app:
...
+
+A Dockerfile to build a personalized image based on a release image
+from GitHub packages, using a custom `bot_config.toml` configuration
+file would look like:
+```dockerfile
+FROM docker.pkg.github.com/coq/bot/coqbot:xxx
+
+COPY bot_config.toml ./
+
+EXPOSE 8000 # The port you specified in bot_config.toml
```
Zimmi48, line 241: I thought we shouldn't specify it if we want to deploy to Heroku...

README.md
```diff
@@ -223,6 +223,26 @@ these are configured in your Heroku app:
...
+set from a configuration file (see [`coqbot.toml`](coqbot.toml)).
+The port number must not be set in the configuration file if you're
+deploying the docker image to Heroku, since it uses a custom
+environment variable.
+
+A Dockerfile to build a personalized image based on a release image
+from GitHub packages, using a custom `bot_config.toml` configuration
+file would look like:
+```dockerfile
+FROM docker.pkg.github.com/coq/bot/coqbot:xxx
```
Zimmi48, line 237: Would using `latest` here work?

Dockerfile
```diff
@@ -25,7 +25,7 @@ FROM alpine:3.10 AS app
 WORKDIR /app
 
 COPY --from=builder /src/opam-install/bin/bot bot.exe
-COPY --from=builder /src/make_ancestor.sh /src/coq_bug_minimizer.sh ./
+COPY --from=builder /src/make_ancestor.sh /src/coq_bug_minimizer.sh /src/coqbot.toml ./
```
Zimmi48, line 28: What is this `/src/coqbot.toml` file? The only one that is defined is `coqbot.toml`, no? And it should be on a layer on top. Maybe this is just because I'm looking at an intermediate step here.

README.md
jtcoolen: Right, the instructions target users which may not want to deploy to Heroku, in which case it is necessary to expose ports.

README.md
jtcoolen: Unfortunately no. But we can replace it with the latest version as of now (v0.1.0). But it wouldn't read the configuration file because this version doesn't support it.

README.md
Zimmi48: OK, maybe this should be made explicit.

Dockerfile
jtcoolen: That's right, this line simplifies to: `COPY make_ancestor.sh coq_bug_minimizer.sh coqbot.toml ./`. There's no need to copy those files from the `builder` stage when it's already accessible.

Zimmi48: Interesting simplification. But this wasn't my main point. My main point was that I would have expected the main `Dockerfile` (the one that is going to get published on GitHub Packages) to not contain any `coqbot.toml` and another `Dockerfile` to be built on top with this configuration file for deployment to Heroku (exactly as documented in the README in fact).

jtcoolen: Oh yeah, this is much cleaner.

jtcoolen: We will need to adapt are current GitHub workflow to deploy to Heroku if we create our docker image from an unpublished release docker image (which must be built beforehand).

.github/workflows/latest_docker_release.yml
```diff
@@ -0,0 +1,19 @@
...
+  build-gh:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: whoan/docker-build-with-cache-action@master
+      with:
+        username: ${{ github.actor }}
+        password: "${{ secrets.GITHUB_TOKEN }}"
+        image_name: jtcoolen/bot/coqbot
+        image_tag: latest
```
Zimmi48, line 16: Unfortunately, from what I understood from [here](https://github.community/t/overwrite-docker-image-in-github-package-registry/16433/10), because GitHub Packages doesn't allow deletion of public packages, the use case of pushing and overwriting the latest tag is not supported.

Dockerfile
```diff
@@ -1,44 +1,5 @@
-FROM ocaml/opam2:alpine-3.10-ocaml-4.10 AS builder
+FROM docker.pkg.github.com/jtcoolen/bot/coqbot:latest
```
Zimmi48, line 1: Instead of pushing the base image to a registry to be able to reuse it on this line, is there a way for this `FROM` directive to use some local image?

.github/workflows/latest_docker_release.yml
jtcoolen: Thanks for mentioning it. Alternatively, we could push to an other registry like Docker Hub from which we can overwrite tagged images ([there's even an action to do that](https://github.com/actions-hub/docker).)

Dockerfile
jtcoolen: Well we could execute a script within our deploy action which would build the release.Dockerfile and tag it (`docker build -f Dockerfile . -t coqbot`), then build our Dockerfile beginning with `FROM coqbot`. The current deploy action looks for a Dockerfile in the repository to build, but I'm not sure if it would find the tagged coqbot image that was just built.

jtcoolen: It turns out that this works pretty well this way.

release.Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+  && apk add bash git \
+  && addgroup coqbot \
+  && adduser coqbot -DG coqbot
+
+COPY --from=builder /src/depexts-coq-bot depexts-coq-bot
+COPY --from=builder /src/depexts-bot-components depexts-bot-components
+
+# Install the required dynamic dependencies
+RUN cat depexts-coq-bot | xargs apk --update add
+RUN cat depexts-bot-components | xargs apk --update add
```
Zimmi48, line 40: It seems to me that you should still keep a default `EXPOSE` and `CMD` here so that the released Docker image is still deployable as is (or else, there is no point on keeping a default bot name in the code, in which case the use of a configuration file would become compulsory).

src/bot.ml
```diff
@@ -130,10 +135,62 @@ let extract_commit json =
...
-let gitlab_ref (issue : GitHub_subscriptions.issue) :
-    GitHub_subscriptions.remote_ref_info =
-  { name= f "pr-%d" issue.number
-  ; repo_url= gitlab_repo ~owner:issue.owner ~name:issue.repo }
+let gitlab_ref ~(issue : GitHub_subscriptions.issue) =
+  let gh_repo = issue.owner ^ "/" ^ issue.repo in
+  let open Lwt.Infix in
+  (* First, we check our hashtable for a key named after the GitHub
+     repository and return the associated GitLab repository. If the
+     key is not found, we load the config file from the default branch.
```
Zimmi48, line 143: We should also document in the README (section "How to use @coqbot instance") about this configuration file.

coqbot.toml
```diff
@@ -0,0 +1,22 @@
...
+#api_token = "xxx"
+
+[mappings]
+  [mappings.coq]
+  github = "coq/coq"
+  gitlab = "coq/coq"
+
+  [mappings.math-comp]
+  github = "math-comp/math-comp"
+  gitlab = "math-comp/math-comp"
```
Zimmi48, line 22: We shouldn't hard-code anything about Math-Comp since this is an independent project and they might want to set it up on their own.

src/bot.ml
```diff
@@ -1024,7 +1108,8 @@ let callback _conn req body =
 
 let server =
   (fun () ->
-    Lwt_io.printf "Initializing repository...\n"
+    Lwt_io.printf "Initializing repository...\n %s\n"
+      (Config.string_of_mapping github_mapping)
```
Zimmi48, line 1112: Not clear to me why this is useful.

coqbot.toml
```diff
@@ -0,0 +1,18 @@
+[bot]
```
Zimmi48, line 1: What about renaming this file to something else than `coqbot.toml` or moving it to a subfolder so that if we ever end up using the bot on this repository, this is not mistaken as the repository-specific configuration file?

coqbot.toml
```diff
@@ -0,0 +1,18 @@
...
+name="coqbot"
+email="coqbot@users.noreply.github.com"
+
+[server]
+domain="coqbot.heroku.com"
+
+[github]
+#webhook_secret = "xxx"
+#api_token = "xxx"
```
Zimmi48, line 10: What about adding a comment explaining why this is commented out?: "In the deployed version, we read these secrets from environment variables."

coqbot.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[gitlab]
+#api_token = "xxx"
+
+[mappings]
+  [mappings.coq]
+  github = "coq/coq"
+  gitlab = "coq/coq"
```
Zimmi48, line 18: If you'd like to add some other mappings to demonstrate how you can do it for several repositories, you can additionally put the mappings for `coq/opam-coq-archive`.

src/bot.ml
```diff
@@ -130,10 +135,62 @@ let extract_commit json =
...
+  ( match gitlab_of_github gh_repo with
+  | None -> (
+      Stdio.printf "No correspondence found for GitHub repository %s/%s.\n"
+        issue.owner issue.repo ;
+      GitHub_queries.get_default_branch ~bot_info ~owner:issue.owner
+        ~repo:issue.repo
+      >>= function
+      | Ok branch -> (
+          GitHub_queries.get_file_content ~bot_info ~owner:issue.owner
+            ~repo:issue.repo ~branch ~file_name:"coqbot.toml"
```
Zimmi48, line 155: What about making the name of this file depend on the name of the bot?

src/bot.ml
```diff
@@ -130,10 +135,62 @@ let extract_commit json =
...
+  | Some r ->
+      Lwt.return r )
+  >|= fun gl_repo ->
+  let owner, name =
+    match Str.split (Str.regexp "/") gl_repo with
+    | [owner; repo] ->
+        (fun () -> Lwt_io.printf "ok %s %s\n" owner repo) |> Lwt.async ;
+        (owner, repo)
+    | _ ->
+        Stdio.printf "Something unexpected happened.\n" ;
```
Zimmi48, line 189: As a general rule, it's best to make these error messages more explicit (even if they should never be raised). Something like "Something unexpected happened during...".

README.md
```diff
@@ -223,6 +233,28 @@ these are configured in your Heroku app:
...
+
+COPY bot_config.toml ./
+
+EXPOSE 8000 # The port you specified in bot_config.toml (this command is ignored if you deploy to Heroku)
```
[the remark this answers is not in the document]
jtcoolen: Because if we comment out the port variable, our instance of the server won't read the port environment variable Heroku is providing. I'm going to document all this and prepare an example configuration file.

coqbot.toml
jtcoolen: We now provide a detailed and commented example configuration file.

PR #81, opened 2020-07-31
Disable merge by email

jtcoolen:
Disable merge by email to guaranty the authenticity of the message.

### DIFF
--- a/bot-components/GitHub_queries.ml
+++ b/bot-components/GitHub_queries.ml
@@ -321,18 +321,12 @@ let pull_request_reviews_info_of_resp ~owner ~repo ~number resp :
                 | Some comments ->
                     comments |> Array.to_list |> List.filter_opt
                     |> List.filter_map ~f:(fun c ->
-                           match
-                             c#author
-                             |> Option.map ~f:(fun (`Actor a) -> a#login)
-                           with
-                           | None ->
-                               None
-                           | Some author ->
-                               Some
-                                 ( { id= c#id
-                                   ; author
-                                   ; created_by_email= c#createdViaEmail }
-                                   : GitHub_subscriptions.comment )) )
+                           c#author
+                           |> Option.map ~f:(fun (`Actor a) ->
+                                  ( { id= c#id
+                                    ; author= a#login
+                                    ; created_by_email= c#createdViaEmail }
+                                    : GitHub_subscriptions.comment ))) )
             ; approved_reviews
             ; comment_reviews=
                 ( match comment_reviews#nodes with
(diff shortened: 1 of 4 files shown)


bot-components/GitHub_queries.ml
```diff
@@ -314,6 +314,25 @@ let pull_request_reviews_info_of_resp ~owner ~repo ~number resp :
...
+                             |> Option.map ~f:(fun (`Actor a) -> a#login)
+                           with
+                           | None ->
+                               None
+                           | Some author ->
+                               Some
+                                 ( { id= c#id
+                                   ; author
+                                   ; created_by_email= c#createdViaEmail }
+                                   : GitHub_subscriptions.comment )) )
```
Zimmi48, line 335: If I'm not mistaken, you can simplify this down to:

```suggestion
                           c#author
                           |> Option.map ~f:(fun (`Actor a) ->
                                  ( { id= c#id
                                    ; author= a#login
                                    ; created_by_email= c#createdViaEmail }
                                    : GitHub_subscriptions.comment ))) )
```

bot.ml
```diff
@@ -839,7 +839,18 @@ let callback _conn req body =
                         ~number:pr.issue.number
                       >>= function
                       | Ok reviews_info -> (
-                          if not (String.equal reviews_info.baseRef "master")
+                          if
+                            List.exists reviews_info.last_comments ~f:(fun c ->
+                                String.equal comment_info.id c.id
+                                && c.created_by_email)
```
Zimmi48, line 845: I think you should change the logic a bit so that one cannot work around this security check by flooding a PR with comments. The logic should say something like: find the comment with the right ID, if not found post an error message about there being too many comments posted since the merge request (should virtually never happen in practice thankfully). If the comment was found, check it was not created by email.

bot.ml
```diff
@@ -839,7 +839,18 @@ let callback _conn req body =
...
+                          then
+                            GitHub_mutations.post_comment ~bot_info
+                              ~message:
+                                (f "@%s: This PR cannot be merged by email."
```
Zimmi48, line 849: I would explain the reason behind this. Something like:

```suggestion
                                (f "@%s: Merge requests sent over e-mail are not \
                                    accepted because this put less guarantee on the \
                                    authenticity of the author of the request."
```

bot.ml
```diff
@@ -839,10 +839,21 @@ let callback _conn req body =
...
                       | Ok reviews_info -> (
-                          if
-                            List.exists reviews_info.last_comments ~f:(fun c ->
-                                String.equal comment_info.id c.id
-                                && c.created_by_email)
+                          let comment =
+                            List.find reviews_info.last_comments ~f:(fun c ->
+                                String.equal comment_info.id c.id)
+                          in
+                          if Option.is_none comment then
```
jtcoolen, line 846: Did this instead of pattern matching in order to avoid one more level of indentation.

Zimmi48: Looks good!

Zimmi48:
This PR introduced a regression: we do not load recent review comments so if the review request was sent through a review comment, then we do not find it.

PR #83, opened 2020-08-06
Refactoring

jtcoolen:
We extract the relevant information from the JSON files only in the `GitHub_subscriptions` and `GitLab_subscriptions` modules, then write interfaces for our modules.


jtcoolen:
In the process I've merged the `/job` and `/pipeline` API endpoints into `/gitlab`, and removed the old `/pull_request` and `/push` as well. This way, only two webhooks need to be set, one for GitHub and the other for GitLab. The status report still works.

We can of course continue to support the legacy endpoints though if you prefer @Zimmi48 .

bot-components/GitLab_subscriptions.ml
```diff
@@ -0,0 +1,117 @@
...
+  {state; id; commit; branch; project_path}
+
+let gitlab_event ~event json =
+  match event with
+  | "Job Hook" ->
+      Ok (`JobEvent (job_info_of_json json))
+  | "Pipeline Hook" ->
+      Ok (`PipelineEvent (pipeline_info_of_json json))
+  | e ->
+      Ok (`UnsupportedEvent e)
```
Zimmi48, line 88: I suggest that you create a datatype for GitLab events, as was done in `GitHub_subscriptions` for GitHub events.

bot-components/GitHub_mutations.mli
```diff
@@ -0,0 +1,42 @@
+val mv_card_to_column :
+     bot_info:Utils.bot_info
+  -> GitHub_queries.mv_card_to_column_input
+  -> unit Lwt.t
+
+val post_comment :
+  bot_info:Utils.bot_info -> id:string -> message:string -> unit Lwt.t
+
+val merge_pull_request :
+     ?merge_method:[< `MERGE | `REBASE | `SQUASH]
```
Zimmi48, line 10: What about providing a dedicated type for this?

bot-components/GitHub_mutations.mli
```diff
@@ -0,0 +1,42 @@
...
+  -> ?commit_body:string
+  -> bot_info:Utils.bot_info
+  -> pr_id:string
+  -> unit Lwt.t
+
+val reflect_pull_request_milestone :
+  bot_info:Utils.bot_info -> GitHub_queries.issue_closer_info -> unit Lwt.t
+
+val add_rebase_label :
+  GitHub_subscriptions.issue -> bot_info:Utils.bot_info -> unit Lwt.t
```
Zimmi48, line 21: Even if this doesn't really change anything, I suggest being consistent and to always put the `bot_info` argument first.

bot-components/GitHub_mutations.mli
```diff
@@ -0,0 +1,42 @@
...
+     repo_full_name:string
+  -> commit:string
+  -> state:string
+  -> url:string
+  -> context:string
+  -> description:string
+  -> bot_info:Utils.bot_info
+  -> unit Lwt.t
+
+val add_pr_to_column : int -> int -> bot_info:Utils.bot_info -> unit Lwt.t
```
Zimmi48, line 42: Since there are two arguments with the same type, we should probably name them to avoid confusion.

bot-components/GitHub_queries.mli
```diff
@@ -0,0 +1,104 @@
...
+type full_backport_info =
+  {backport_info: backport_info list; rejected_milestone: string}
+
+val get_backport_info :
+  bot_info:Utils.bot_info -> string -> full_backport_info option
+
+type project_card =
+  { id: string
+  ; column: GitHub_GraphQL.project_column option
+  ; columns: GitHub_GraphQL.project_column list }
```
Zimmi48, line 21: We should not export anything from this internal module (`GitHub_GraphQL`). In fact, I introduced it only because I couldn't make Merlin understand the ppx but since it works fine now, there is no reason to keep this module anymore and we could move each GraphQL request next to the functions processing them.

bot-components/GitHub_queries.mli
```diff
@@ -0,0 +1,104 @@
...
+  -> team:string
+  -> user:string
+  -> (bool, string) result Lwt.t
+
+val get_pull_request_refs :
+     bot_info:Utils.bot_info
+  -> owner:string
+  -> repo:string
+  -> number:int
+  -> (string GitHub_subscriptions.pull_request_info, string) result Lwt.t
```
Zimmi48, line 57: Instead of having some types defined in `GitHub_subscriptions`, some defined in `GitHub_queries` and some in `GitHub_mutations`, we could consider introducing a new `GitHub_types` module to gather them. It would also present the advantage of not requiring to duplicate the type definition with a `modules_without_implementation`. See https://dune.readthedocs.io/en/stable/dune-files.html#library

bot-components/GitHub_mutations.mli
```diff
@@ -0,0 +1,42 @@
+val mv_card_to_column :
+     bot_info:Utils.bot_info
+  -> GitHub_queries.mv_card_to_column_input
+  -> unit Lwt.t
+
+val post_comment :
+  bot_info:Utils.bot_info -> id:string -> message:string -> unit Lwt.t
```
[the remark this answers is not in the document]
Zimmi48: I'm talking about the object identifiers that are necessary for some requests in the GraphQL API.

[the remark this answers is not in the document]
jtcoolen: Like wrapping in this example the strings `id` and `message` in a dedicated type for GraphQL objects ids, right?

bot-components/GitHub_subscriptions.mli
```diff
@@ -0,0 +1,86 @@
...
+  | IssueClosed of issue_info
+  | RemovedFromProject of project_card
+  | PullRequestUpdated of pull_request_action * issue_info pull_request_info
+  | BranchCreated of remote_ref_info
+  | TagCreated of remote_ref_info
+  | CommentCreated of comment_info
+  | CheckRunCreated of check_run_info
+  | PushEvent of push_info
+
+val issue_info_of_json :
```
Zimmi48, line 69: If I'm not mistaken, none of the `_of_json` functions are used outside this file. They should be considered internal and not exported.

bot-components/GitHub_mutations.mli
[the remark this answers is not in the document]
Zimmi48: Only `id`. And you don't actually have to wrap the type. You can define a type alias (`type id = string`) in the ml file and make it abstract (`type id`) in the mli files.

bot-components/GitHub_mutations.ml
```diff
@@ -32,6 +32,8 @@ let update_milestone ~bot_info ~issue ~milestone =
   | Error err ->
       Stdio.print_endline (f "Error while updating milestone: %s" err)
 
+type merge_method = [`MERGE | `REBASE | `SQUASH]
```
Zimmi48, line 35: I meant define a type with normal constructors (and convert internally those constructors into the required variants).

bot-components/GitLab_subscriptions.ml
```diff
@@ -0,0 +1,122 @@
...
+  | "Job Hook" ->
+      Ok (JobEvent (job_info_of_json json))
+  | "Pipeline Hook" ->
+      Ok (PipelineEvent (pipeline_info_of_json json))
+  | e ->
+      Ok (UnsupportedEvent e)
+
+let receive_gitlab ~secret headers body =
+  let open Result in
+  ( match Header.get headers "X-Hub-Signature" with
```
Zimmi48, line 97: Is this really the name of the header? This looks like a copy-paste typo.

bot-components/GitLab_subscriptions.ml
```diff
@@ -0,0 +1,122 @@
...
+  | Some signature ->
+      let expected =
+        Nocrypto.Hash.SHA1.hmac ~key:(Cstruct.of_string secret)
+          (Cstruct.of_string body)
+        |> Hex.of_cstruct |> Hex.show |> f "sha1=%s"
+      in
+      (* TODO: move to constant time equality function, such as https://github.com/mirage/eqaf,
```
Zimmi48, line 104: You can fix this now.

bot-components/GitLab_subscriptions.ml
```diff
@@ -0,0 +1,122 @@
...
+    try
+      let json = Yojson.Basic.from_string body in
+      gitlab_event ~event json |> Result.map ~f:(fun r -> (signed, r))
+    with
+    | Yojson.Json_error err ->
+        Error (f "Json error: %s" err)
+    | Type_error (err, _) ->
+        Error (f "Json type error: %s" err) )
+  | None ->
+      Error "Not a GitHub webhook."
```
Zimmi48, line 122: Another copy-paste typo.

bot-components/GitLab_subscriptions.mli
```diff
@@ -0,0 +1,27 @@
...
+  ; failure_reason: string option
+  ; allow_fail: bool option }
+
+type pipeline_info =
+  {state: string; id: int; commit: string; branch: string; project_path: string}
+
+type msg =
+  | JobEvent of job_info
+  | PipelineEvent of pipeline_info
+  | UnsupportedEvent of string
```
Zimmi48, line 18: This is a better name than the `NoOp` from `GitHub_subscriptions` so I suggest adopting this name in both files.

bot-components/GitLab_subscriptions.mli
```diff
@@ -0,0 +1,27 @@
...
+
+val pr_from_branch : string -> int option * string
```
Zimmi48, line 20: This function and the two below sound internal to me and thus shouldn't be exported.

bot-components/GitHub_mutations.mli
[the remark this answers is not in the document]
jtcoolen: Done! We may want to rename some `id` integer fields in some records, which refer to the REST API. I'll look into it.

bot-components/GitHub_queries.mli
jtcoolen: Why not make it a private module? It makes sense to separate the GraphQL requests from the functions processing. I'm also afraid that the file might get too long.

bot-components/GitLab_subscriptions.ml
jtcoolen: No it's not! A request header from GitLab contains three fields: Content-Type, X-Gitlab-Event and X-Gitlab-Token (plaintext).

jtcoolen: Should we ignore GitLab payload with the wrong webhook secret? Should we also keep the same password for GitHub and GitLab webhook payloads?

jtcoolen: We should check if the payloads coming from GitHub and GitLab are formatted in JSON by reading the `content-type` header too, right?

Zimmi48: > Should we ignore GitLab payload with the wrong webhook secret?

The best thing to do would be to reject them by replying with a 401 error code and logging the failed authentication attempt in the log.

>Should we also keep the same password for GitHub and GitLab webhook payloads?

It would be best not to. If we want to generalize the GitLab webhook secrets, we may eventually need to generate one secret per project as we discussed.

> We should check if the payloads coming from GitHub and GitLab are formatted in JSON by reading the `content-type` header too, right?

Yes, that would be a good thing to do.

bot-components/GitHub_queries.mli
Zimmi48: That's fine by me.

bot-components/GitHub_types.mli
```diff
@@ -0,0 +1,95 @@
+type id = string
```
Zimmi48, line 1: The only problem with making `GitHub_types` an mli-only module is that we cannot abstract the `id` type (and if we don't, providing a type alias doesn't bring much). So in the end, we need to provide an ml implementation for this file or move the `id` type definition to another module.

bot-components/GitHub_subscriptions.mli
```diff
@@ -0,0 +1,5 @@
+val receive_github :
+     secret:string
+  -> Cohttp.Header.t
+  -> string
+  -> (bool * GitHub_types.msg, string) result
```
Zimmi48, line 5: The one type I would keep in this module instead of moving it elsewhere is the `msg` type. It is not really a type that may need to be shared between `GitHub_subscriptions`, `GitHub_queries` and `GitHub_mutations`.

bot-components/GitHub_queries.mli
```diff
@@ -0,0 +1,104 @@
...
+  -> ('a, string) result Lwt.t
+
+type backport_info =
+  {backport_to: string; request_inclusion_column: int; backported_column: int}
+
+type full_backport_info =
+  {backport_info: backport_info list; rejected_milestone: string}
+
+val get_backport_info :
+  bot_info:Utils.bot_info -> string -> full_backport_info option
```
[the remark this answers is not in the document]
jtcoolen: This function doesn't return a Lwt.t because it doesn't make any calls to GraphQL or Rest queries. So I moved it to the Bot_components.Utils module since it's not a query: it parses a string containing some backport information.

bot-components/Bot_info.mli
```diff
@@ -0,0 +1,6 @@
+type bot_info =
```
Zimmi48, line 1: If you create a module dedicated to a type, the OCaml convention is then to call this type `t`.

bot-components/GitHub_queries.mli
```diff
@@ -0,0 +1,74 @@
+val pull_request_milestone_and_cards :
+     bot_info:Bot_info.bot_info
+  -> owner:string
+  -> repo:string
+  -> number:int
+  -> ( GitHub_types.project_card list * GitHub_GraphQL.milestone option
```
Zimmi48, line 6: Here we still use a type coming from `GitHub_GraphQL`. I don't know how this can work since this module is supposed to be private, but the type definitions from `GitHub_GraphQL` should be moved to `GitHub_types`.

bot-components/Utils.mli
```diff
@@ -0,0 +1,28 @@
...
+val print_response : Cohttp.Response.t * Cohttp_lwt.Body.t -> unit Lwt.t
+
+val send_request :
+     bot_info:Bot_info.bot_info
+  -> body:Cohttp_lwt.Body.t
+  -> uri:Uri.t
+  -> (string * string) list
+  -> unit Lwt.t
+
+val get_backport_info :
```
Zimmi48, line 17: This function is only used in `GitHub_queries` as far as I can see. So instead of moving it to `Utils`, you could simply have not exported it (and also possibly renamed it so that its name doesn't suggest that it is one of the functions that is exported).

bot-components/GitHub_subscriptions.mli
```diff
@@ -0,0 +1,16 @@
+type msg =
+  | IssueOpened of GitHub_types.issue_info
```
Zimmi48, line 2: Same remark on `open GitHub_types`.

bot-components/GitLab_subscriptions.mli
```diff
@@ -0,0 +1,7 @@
+type msg =
+  | JobEvent of GitLab_types.job_info
```
Zimmi48, line 2: Same remark regarding `open GitLab_types`.

src/actions.mli
```diff
@@ -0,0 +1,88 @@
...
+     bot_info:Bot_components.Bot_info.t
+  -> string
+  -> (Cohttp.Response.t * Cohttp_lwt__Body.t) Lwt.t
+
+val merge_pull_request :
+     bot_info:Bot_components.Bot_info.t
+  -> comment_info:Bot_components.GitHub_types.comment_info
+  -> unit Lwt.t
+
+val pull_request_updated :
```
Zimmi48, line 23: If I'm not mistaken, `pull_request_updated` is not used outside this module so should not be exported (the function that is called from outside being `pull_request_updated_action` or `run_ci`).

src/actions.mli
```diff
@@ -0,0 +1,88 @@
...
+val run_ci :
+     bot_info:Bot_components.Bot_info.t
+  -> comment_info:Bot_components.GitHub_types.comment_info
+  -> gitlab_mapping:(string, string) Base.Hashtbl.t
+  -> github_mapping:(string, string) Base.Hashtbl.t
+  -> gitlab_of_github:(string -> string option)
+  -> signed:bool
+  -> (Cohttp.Response.t * Cohttp_lwt__Body.t) Lwt.t
+
+val pull_request_closed :
```
Zimmi48, line 42: For consistency, this one should probably be renamed `pull_request_closed_action`.

bot-components/GitHub_queries.ml
```diff
@@ -131,17 +93,16 @@ let pull_request_milestone_and_cards ~bot_info ~owner ~repo ~number =
...
 
-type mv_card_to_column_input = {card_id: string; column_id: string}
-
-let backported_pr_info ~bot_info number base_ref =
-  pull_request_milestone_and_cards ~bot_info ~owner:"coq" ~repo:"coq" ~number
-  >|= function
+let get_backported_pr_info ~bot_info number base_ref =
+  get_pull_request_milestone_and_cards ~bot_info ~owner:"coq" ~repo:"coq"
+    ~number
+  >>= function
```
Zimmi48, line 99: Instead of a `>>=` here and a `Lwt.return` in each branch, you can probably use a `>|=` here.

src/actions.mli
```diff
@@ -0,0 +1,88 @@
...
+  -> base_ref:string
+  -> commits_msg:string list
+  -> unit
+  -> unit Lwt.t
+
+val repeat_request : string Lwt.t -> string Lwt.t
+
+type build_failure = Warn | Retry | Ignore
+
+val trace_action : repo_full_name:string -> string -> build_failure
```
Zimmi48, line 88: As far as I can tell, this type and these two functions are purely internal to the `Actions` module and should not be exported...

src/actions.ml
```diff
@@ -0,0 +1,782 @@
...
+    request
+    >>= fun body ->
+    if String.is_empty body then Lwt_unix.sleep t >>= fun () -> aux (t *. 2.)
+    else Lwt.return body
+  in
+  aux 2.
+
+type build_failure = Warn | Retry | Ignore
+
+let trace_action ~repo_full_name trace =
```
Zimmi48, line 732: This is used in a commented code in the `job_failure` function, so this should be moved above.

src/bot.ml
```diff
@@ -670,452 +89,101 @@ let callback _conn req body =
...
-        |&& f "git config user.email \"%s\"" bot_email
-        |&& f "git config user.name \"%s\"" bot_name
-        |> execute_cmd >|= ignore ))
-  |> Lwt.async ;
+let launch =
+  init_git_bare_repository ~bot_info
+  >>= fun () ->
+  Stdio.printf "Starting server.\n" ;
   let mode = `TCP (`Port port) in
   Server.create ~mode (Server.make ~callback ())
```
Zimmi48, line 183: I'm a bit worried that the change of strategy here (wait until the git repository is initialized before creating the server) is going to generate even more timeouts.

src/config.ml
```diff
@@ -46,6 +46,13 @@ let github_webhook_secret toml_data =
   | Some secret ->
       secret
 
+let gitlab_webhook_secret toml_data =
+  match subkey_value toml_data "gitlab" "webhook_secret" with
+  | None ->
+      Sys.getenv_exn "GITLAB_WEBHOOK_SECRET"
+  | Some secret ->
+      secret
```
Zimmi48, line 54: Since we will probably need to refine this anyway, let's make it optional in the meantime. I.e., if `GITLAB_WEBHOOK_SECRET` is not defined, we could default to, say `GITHUB_WEBHOOK_SECRET`.

src/git_utils.mli
```diff
@@ -0,0 +1,47 @@
+val gitlab_repo :
```
Zimmi48, line 1: This is very likely not used anywhere outside this module, so should not be exported.

src/git_utils.mli
```diff
@@ -0,0 +1,47 @@
...
+  bot_info:Bot_components.Bot_info.t -> owner:string -> name:string -> string
+
+val report_status : string -> string -> int -> ('a, string) result
```
Zimmi48, line 4: Same thing.

src/git_utils.mli
```diff
@@ -0,0 +1,47 @@
...
+  -> remote_ref:Bot_components.GitHub_types.remote_ref_info
+  -> local_ref:string
+  -> string
+
+val git_delete :
+  remote_ref:Bot_components.GitHub_types.remote_ref_info -> string
+
+val git_make_ancestor : base:string -> string -> (bool, string) result Lwt.t
+
+val git_coq_bug_minimizer :
```
Zimmi48, line 32: As far as I can tell, this is an auxiliary function for `run_coq_minimizer` so this one should probably not be exported.

src/bot.ml
jtcoolen: Now we initialize the git bare repository before each action calling git commands under the hood.

PR #84, opened 2020-08-10
Constant time equality

jtcoolen:
Let's compare cryptographic hashes in constant time to prevent timing attacks from happening.
These changes depend on #83


Zimmi48:
LGTM. It seems to me that the dependency order between this PR and #83 could be inverted so that this one is merged now.

Zimmi48:
There were some build issues when trying to deploy this. First, the eqaf dependency had been added to the wrong package. I've fixed this. Second, the library should be listed in the Dune file. I'm on my phone now so cannot test this. I don't understand why it didn't fail while building in a nix-shell though.

jtcoolen:
Thanks for fixing what I overlooked.

PR #86, opened 2020-08-18
Authenticate as GitHub app

jtcoolen:


src/bot.ml
```diff
@@ -28,12 +28,52 @@ let bot_info : Bot_components.Bot_info.t =
...
+        ( Stdio.printf "No correspondence found for GitLab repository %s.\n"
+            gitlab_full_name ;
+          gitlab_full_name )
+  in
+  match Str.split (Str.regexp "/") repo_full_name with
+  | [owner_; repo_] ->
+      (owner_, repo_)
+  | _ ->
+      (* Shouldn't happen *)
+      ("owner", "repo")
```
Zimmi48, line 65: In this case, you could simply raise an exception (with `failwith`).

src/actions.ml
```diff
@@ -320,21 +309,29 @@ let pipeline_action ~bot_info pipeline_info ~github_of_gitlab : unit Lwt.t =
...
+                       bot_info.name coq_minimizer_repo_token repo_name
+                       branch_name)
+                >>= function
+                | Ok () ->
+                    Lwt.return ()
+                | Error f ->
+                    Lwt_io.printf "Error: %s" f )
+          in
+          action_as_github_app ~bot_info ~key ~app_id ~owner ~repo
+            ~installation_tokens post_results_remove_ci_branch)
```
Zimmi48, line 334: What about making it clearer that the "remove CI branch" part is using a different token by moving the `<&> ...` part out of the `action_as_github_app` call? This would be equivalent, right?

src/bot.ml
```diff
@@ -28,12 +28,23 @@ let bot_info : Bot_components.Bot_info.t =
...
+let app_id = Config.github_app_id toml_data
+
 let github_mapping, gitlab_mapping = Config.make_mappings_table toml_data
 
 let github_of_gitlab = Hashtbl.find gitlab_mapping
 
 let gitlab_of_github = Hashtbl.find github_mapping
 
+let installation_tokens : (string, string * float) Base.Hashtbl.t =
+  Hashtbl.create (module String)
```
Zimmi48, line 42: Wouldn't it make sense to move this definition to `Helpers` (or an entirely new module exporting only the `action_as_github_app` function)? This way, it wouldn't be necessary to pass this variable as a parameter at every `action_as_github_app` call.

src/config.ml
```diff
@@ -72,6 +72,31 @@ let bot_email toml_data =
...
+  | Some app_id ->
+      app_id |> Int.of_string
+
+(*let string_of_file_path path = Stdio.In_channel.(with_file path ~f:input_all)*)
+
+let github_private_key =
+  (*string_of_file_path "./github.private-key.pem"*)
+  match
+    let private_k = Sys.getenv_exn "GITHUB_PRIVATE_KEY" in
+    Stdio.eprintf "Found private key: %s\n" private_k ;
```
Zimmi48, line 90: Why `eprintf` and not `printf` in this case?

src/helpers.ml
```diff
@@ -33,3 +35,61 @@ let trim_comments comment =
...
+      action ~bot_info
+
+let action_as_github_app ~bot_info ~key ~app_id ~owner ~repo
+    ~installation_tokens action =
+  (* Executes an action with an installation token if the repository has
+     the GitHub app installed.
+     Generates a new installation token if the existing one has expired. *)
+  match Hashtbl.find installation_tokens owner with
+  | Some (github_token, expiration_date) ->
+      if Float.(expiration_date < Unix.time ()) then (
```
Zimmi48, line 60: I thought you wanted to take a 20 minutes margin. You removed it from `GitHub_app` without reintroducing it here.

src/helpers.ml
```diff
@@ -33,3 +35,61 @@ let trim_comments comment =
...
+  GitHub_app.get_installation_token ~bot_info ~key ~app_id ~owner ~repo
+  >>= function
+  | Ok (github_token, expiration_date) ->
+      let _ =
+        Hashtbl.add installation_tokens ~key:owner
+          ~data:(github_token, expiration_date)
+      in
+      let bot_info : Bot_info.t = {bot_info with github_token} in
+      action ~bot_info
+  | Error _ ->
```
Zimmi48, line 50: Deserves a comment explaining why we use this fallback for now.

Zimmi48 (commented):
If you do apply some of my suggestions, make sure to properly test the result to make sure that nothing is broken.

src/helpers.ml
jtcoolen: The 20 minutes margin is still in `GitHub_app,` no? The part where we set the expiration date.

Zimmi48: Oh right, I misread.

README.md
```diff
@@ -125,7 +125,10 @@ you are interested in hearing about it.
 
 ## How to use the **@coqbot** instance ##
 
-To use the bot without deploying your own instance, follow these steps:
+### As a GitHub App
+The bot can be installed as a GitHub App to either your account or
+organization (insert link to the app in the marketplace). Once you
```
Zimmi48, line 130: It's not listed (yet) in the marketplace but there is already a public installation link: https://github.com/apps/coqbot-app

README.md
```diff
@@ -150,6 +153,28 @@ To use the bot without deploying your own instance, follow these steps:
     <https://coqbot.herokuapp.com/job> that will only be triggered by
     job events.
```
Zimmi48, line 154: This step can now be simplified: the user can create a single webhook (`/gitlab`) triggering both pipeline and job events.

README.md
```diff
@@ -125,7 +125,10 @@ you are interested in hearing about it.
 
 ## How to use the **@coqbot** instance ##
 
-To use the bot without deploying your own instance, follow these steps:
+### As a GitHub App
```
Zimmi48, line 128: Please add a warning that installing as a GitHub App is still in the experimental stage and that users may frequently receive requests to expand permissions.

Zimmi48 (approved):

PR #93, opened 2020-08-25
Checks API

jtcoolen:
Preliminary support for the GitHub Checks API.

### DIFF
--- a/bot-components/GitHub_GraphQL.ml
+++ b/bot-components/GitHub_GraphQL.ml
@@ -191,6 +191,16 @@ module FileContent =
   }
 |}]
 
+module RepoId =
+[%graphql
+{|
+  query repoId($owner: String!, $repo: String!) {
+    repository(owner: $owner, name: $repo) {
+      id
+    }
+  }
+|}]
+
 (* Mutations *)
 
 module MoveCardToColumn =
@@ -234,6 +244,45 @@ module MergePullRequest =
         state
         url
       }
+    }
+  }
+|}]
+
+module NewCheckRun =
+[%graphql
+{|
+  mutation newCheckRun($name: String!, $repoId: ID!, $headSha: String!, $status: String!, $title: String!, $text: String!) {
+    createCheckRun(
+      input: {
+        status:$status,
+        name:$name,
+        repositoryId:$repoId,
+        headSha:$headSha,
+        output:{
+          title:$title,
+          text:$text
+        }
+      }) {
+      clientMutationId
+    }
+  }
+|}]
+
+module UpdateCheckRun =
+[%graphql
+{|
+  mutation updateCheckRun($checkRunId: ID!, $repoId: ID!, $conclusion: String!, $title: String!, $text: String!) {
+    updateCheckRun(
+      input: {
+        checkRunId:$checkRunId,
+        repositoryId:$repoId,
+        conclusion:$conclusion,
+        output:{
+          title:$title,
+          text:$text,
+        }
+      }) {
+      clientMutationId
     }
   }
 |}]
--- a/bot-components/GitHub_GraphQL.ml
+++ b/bot-components/GitHub_GraphQL.ml
@@ -236,8 +236,11 @@ module UpdateMilestone =
 module MergePullRequest =
 [%graphql
 {|
-  mutation mergePullRequest($pr_id: ID!, $commit_headline: String, $commit_body: String, $merge_method: PullRequestMergeMethod) {
-    mergePullRequest(input: {pullRequestId: $pr_id, commitHeadline: $commit_headline, commitBody: $commit_body, mergeMethod: $merge_method}) {
+  mutation mergePullRequest($pr_id: ID!, $commit_headline: String,
+  $commit_body: String, $merge_method: PullRequestMergeMethod) {
+    mergePullRequest(
+      input: {pullRequestId: $pr_id, commitHeadline: $commit_headline,
+      commitBody: $commit_body, mergeMethod: $merge_method}) {
       pullRequest {
         merged
         mergedAt
@@ -251,13 +254,16 @@ module MergePullRequest =
 module NewCheckRun =
 [%graphql
 {|
-  mutation newCheckRun($name: String!, $repoId: ID!, $headSha: String!, $status: String!, $title: String!, $text: String!, $summary: String!) {
+  mutation newCheckRun($name: String!, $repoId: ID!, $headSha: String!,
+  $status: String!, $title: String!, $text: String!, $summary: String!,
+  $conclusion: String!) {
     createCheckRun(
       input: {
         status:$status,
         name:$name,
         repositoryId:$repoId,
         headSha:$headSha,
+        conclusion:$conclusion,
         output:{
           title:$title,
           text:$text,
@@ -272,7 +278,9 @@ module NewCheckRun =
 module UpdateCheckRun =
 [%graphql
 {|
-  mutation updateCheckRun($checkRunId: ID!, $repoId: ID!, $conclusion: String!, $title: String!, $text: String!, $summary: String!) {
+  mutation updateCheckRun($checkRunId: ID!, $repoId: ID!
+  $conclusion: String!, $title: String!, $text: String!,
+  $summary: String!) {
     updateCheckRun(
       input: {
         checkRunId:$checkRunId,
--- a/bot-components/GitHub_GraphQL.ml
+++ b/bot-components/GitHub_GraphQL.ml
@@ -251,7 +251,7 @@ module MergePullRequest =
 module NewCheckRun =
 [%graphql
 {|
-  mutation newCheckRun($name: String!, $repoId: ID!, $headSha: String!, $status: String!, $title: String!, $text: String!) {
+  mutation newCheckRun($name: String!, $repoId: ID!, $headSha: String!, $status: String!, $title: String!, $text: String!, $summary: String!) {
     createCheckRun(
       input: {
         status:$status,
@@ -260,7 +260,8 @@ module NewCheckRun =
         headSha:$headSha,
         output:{
           title:$title,
-          text:$text
+          text:$text,
+          summary:$summary
         }
       }) {
       clientMutationId
@@ -271,7 +272,7 @@ module NewCheckRun =
 module UpdateCheckRun =
 [%graphql
 {|
-  mutation updateCheckRun($checkRunId: ID!, $repoId: ID!, $conclusion: String!, $title: String!, $text: String!) {
+  mutation updateCheckRun($checkRunId: ID!, $repoId: ID!, $conclusion: String!, $title: String!, $text: String!, $summary: String!) {
     updateCheckRun(
       input: {
         checkRunId:$checkRunId,
@@ -280,6 +281,7 @@ module UpdateCheckRun =
         output:{
           title:$title,
           text:$text,
+          summary:$summary
         }
       }) {
       clientMutationId
--- a/bot-components/GitHub_GraphQL.ml
+++ b/bot-components/GitHub_GraphQL.ml
@@ -256,7 +256,7 @@ module NewCheckRun =
 {|
   mutation newCheckRun($name: String!, $repoId: ID!, $headSha: String!,
   $status: String!, $title: String!, $text: String!, $summary: String!,
-  $conclusion: String!) {
+  $conclusion: String) {
     createCheckRun(
       input: {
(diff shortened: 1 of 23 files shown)


bot-components/GitHub_queries.ml
```diff
@@ -415,6 +415,20 @@ let get_issue_closer_info ~bot_info ({owner; repo; number} : issue) =
...
+  | None ->
+      Error (f "Unknown repository %s/%s." owner repo)
+  | Some repository ->
+      Ok repository#id
+
+let get_repository_id ~bot_info ~owner ~repo =
+  RepoId.make ~owner ~repo ()
+  |> GraphQL_query.send_graphql_query ~bot_info
+  >|= Result.map_error ~f:(fun err ->
+          f "Query get_default_branch failed with %s" err)
```
Zimmi48, line 429: Copy-paste mistake.

bot-components/GitHub_GraphQL.ml
```diff
@@ -237,3 +250,49 @@ module MergePullRequest =
...
+        status:$status,
+        name:$name,
+        repositoryId:$repoId,
+        headSha:$headSha,
+        conclusion:$conclusion,
+        output:{
+          title:$title,
+          text:$text,
+          summary:$summary
+        }
```
Zimmi48, line 271: You can (and you shoud) add the `detailsUrl` argument so that we can provide a different details URL than the default. (Cf. https://docs.github.com/en/graphql/reference/input-objects#createcheckruninput.)

bot-components/GitHub_GraphQL.ml
```diff
@@ -237,3 +250,49 @@ module MergePullRequest =
...
+    updateCheckRun(
+      input: {
+        checkRunId:$checkRunId,
+        repositoryId:$repoId,
+        conclusion:$conclusion,
+        output:{
+          title:$title,
+          text:$text,
+          summary:$summary
+        }
```
Zimmi48, line 293: Same thing. Cf. https://docs.github.com/en/graphql/reference/input-objects#updatecheckruninput

bot-components/GitHub_mutations.ml
```diff
@@ -71,6 +71,62 @@ let reflect_pull_request_milestone ~bot_info issue_closer_info =
...
+  | SUCCESS ->
+      "SUCCESS"
+  | TIMED_OUT ->
+      "TIMED_OUT"
+
+let create_check_run ~bot_info ?conclusion ~name ~repo_id ~head_sha ~status
+    ~title ~text ~summary =
+  let conclusion =
+    match conclusion with None -> "" | Some c -> string_of_conclusion c
+  in
```
Zimmi48, line 97: I'm not sure I understand the logic around `conclusion`. If this is an optional argument, can't you keep it optional all the way to the GraphQL request (remove the `!`) instead of having to convert `None` to `""` (does this even work)?

bot-components/GitHub_queries.ml
```diff
@@ -443,3 +457,37 @@ let get_cards_in_column column_id =
...
              else None))
+
+let get_installations =
+  generic_get "installation/repositories" ~header_list:app_api_preview_header
+    (fun json ->
+      let open Yojson.Basic.Util in
+      json |> member "repositories" |> to_list
+      |> List.map ~f:(fun json ->
+             ( json |> member "owner" |> member "login" |> to_string
+             , json |> member "name" |> to_string )))
```
Zimmi48, line 468: What is this doing in this PR? Has this escaped from the previous version of #100?

bot-components/GitHub_types.mli
```diff
@@ -81,7 +83,39 @@ type comment_info =
...
+
+type check_run =
+  { id: int
+  ; node_id: id
+  ; head_sha: string
+  ; name: string
+  ; status: string
+  ; url: string
+  ; title: string
+  ; text: string }
```
Zimmi48, line 121: What is the reason for having these two types? Where is `check_run` used?

bot-components/GitHub_queries.ml
jtcoolen: I messed up while rebasing.

bot-components/GitHub_types.mli
jtcoolen: The `check_run` record is used in the `get_check_runs` query. However, a simplification in the code make it no longer relevant. We can remove it if you prefer.

src/actions.ml
```diff
@@ -30,21 +35,34 @@ let send_status_check ~bot_info job_info pr_num (gh_owner, gh_repo)
...
+                (first_line_of_string commit_message)
+                ( if
+                  String.equal job_info.build_name
+                    "library:ci-fiat_crypto_legacy"
+                then "\nping @JasonGross"
+                else "" )
+            in
+            match bot_info.github_token with
+            | ACCESS_TOKEN _t ->
+                GitHub_mutations.post_comment ~bot_info ~id ~message
```
Zimmi48, line 53: I don't think we should keep the comment back-up mode. Since reporting allow failure mode jobs is a new feature that is only used in Coq (as far as I can tell), it seems acceptable to me to restrict this feature to GitHub Apps only.

src/actions.ml
```diff
@@ -30,21 +35,34 @@ let send_status_check ~bot_info job_info pr_num (gh_owner, gh_repo)
...
-                    mode has failed for commit %s: %s%s"
-                   job_info.build_name repo_full_name job_info.build_id
-                   job_info.commit
-                   (first_line_of_string commit_message)
-                   ( if
-                     String.equal job_info.build_name
-                       "library:ci-fiat_crypto_legacy"
-                   then "\nping @JasonGross"
-                   else "" ))
+          when String.equal head.sha (f "\"%s\"" job_info.commit) -> (
```
Zimmi48, line 38: Since we're moving to checks, the logic for checking whether we are in a PR and on the latest commit is no longer relevant and can be removed (we can post checks in all cases).

src/actions.ml
```diff
@@ -30,21 +35,34 @@ let send_status_check ~bot_info job_info pr_num (gh_owner, gh_repo)
...
+            let message =
+              f
+                "For your complete information, the job [%s](%s) in allow \
+                 failure mode has failed for commit %s: %s%s"
+                job_info.build_name job_url job_info.commit
+                (first_line_of_string commit_message)
```
Zimmi48, line 44: Since this is going to be displayed in the Checks tab, the message should be adjusted. In particular, there is no need to adopt a discussion-like language ("For your complete information") nor is there to specify for which commit the build has failed.

src/actions.ml
```diff
@@ -30,21 +35,34 @@ let send_status_check ~bot_info job_info pr_num (gh_owner, gh_repo)
...
+                ( if
+                  String.equal job_info.build_name
+                    "library:ci-fiat_crypto_legacy"
+                then "\nping @JasonGross"
+                else "" )
```
Zimmi48, line 49: The feature for pinging Jason is still relevant AFAICT and should still be done as a comment in all cases.

src/actions.ml
```diff
@@ -61,40 +79,78 @@ let send_status_check ~bot_info job_info pr_num (gh_owner, gh_repo)
...
+    | INSTALL_TOKEN _t -> (
+        GitHub_queries.get_repository_id ~bot_info ~owner:gh_owner ~repo:gh_repo
+        >>= function
+        | Ok repo_id ->
+            GitHub_mutations.create_check_run ~bot_info ~name:context ~repo_id
+              ~head_sha:job_info.commit ~conclusion:FAILURE ~status:COMPLETED
+              ~title:(failure_reason ^ " on GitLab CI")
+              ~text:job_url ~details_url:job_url ~summary:job_url
+        | Error _ ->
+            Lwt.return () )
```
Zimmi48, line 97: It seems to me that the exact same logic is repeated often, so it could be convenient to move it to an external function and reuse it where convenient (not necessarily in places where there are small variations / deviations from this exact code / logic).

Regarding the last two lines, why not print something in the log when that happens? (In some other places, you already do it BTW.)

src/actions.ml
```diff
@@ -265,49 +339,102 @@ let job_action ~bot_info (job_info : job_info) ~github_of_gitlab =
...
+                let conclusion, title =
+                  match pipeline_info.state with
+                  | "success" ->
+                      (SUCCESS, "Pipeline completed on GitLab CI")
+                  | "failed" ->
+                      (FAILURE, "Pipeline completed with errors on GitLab CI")
+                  | "cancelled" ->
+                      (CANCELLED, "Pipeline was cancelled on GitLab CI")
+                  | s ->
+                      (ACTION_REQUIRED, "Unknown pipeline status: " ^ s)
```
Zimmi48, line 427: I'm not convinced by this mapping. I would still use `FAILURE` in this case.

src/actions.ml
```diff
@@ -540,16 +668,31 @@ let update_pr ~bot_info pr_info ~gitlab_mapping ~github_mapping
...
+          ~description:
+            "Pipeline did not run on GitLab CI because PR has conflicts with \
+             base branch."
+          ~bot_info
+        |> Lwt_result.ok
+    | INSTALL_TOKEN _t -> (
+        GitHub_queries.get_repository_id ~bot_info
+          ~owner:pr_info.issue.issue.owner ~repo:pr_info.issue.issue.repo
+        >>= function
+        | repo_id ->
```
Zimmi48, line 687: Why don't you distinguish between `Ok` and `Error` here? Does the code even compile?

BTW, in this case it might be possible to get the repo id from the webhook payload and not have to do a query to get it (no need to take care of this optimization if you don't want to / don't have time).

Zimmi48 (commented):
Looks generally good to me. A few comments though.

bot-components/GitHub_types.mli
Zimmi48: No need.

src/actions.ml
jtcoolen: The bind function from Lwt_result is used instead of the one from Lwt, due to the local open of `Lwt_result.Infix`.

src/actions.ml
```diff
@@ -28,28 +41,37 @@ let send_status_check ~bot_info job_info pr_num (gh_owner, gh_repo)
...
+                      ~title:(failure_reason ^ " on GitLab CI")
+                      ~text:"" ~details_url:job_url ~summary:message
+                | Error e ->
+                    Lwt_io.printf "No repo id: %s\n" e )
+              ~github_access_token_func:
+                ( if
+                  String.equal job_info.build_name
+                    "library:ci-fiat_crypto_legacy"
+                then GitHub_mutations.post_comment ~bot_info ~id ~message
+                else Lwt.return () )
```
Zimmi48, line 74: This is executed only in the `github_access_token` case when it is in fact something very specific to the Coq repository itself, which uses the GitHub App.

src/actions.ml
```diff
@@ -28,28 +33,39 @@ let send_status_check ~bot_info job_info pr_num (gh_owner, gh_repo)
...
+                job_info.build_name job_url
+                ( if
+                  String.equal job_info.build_name
+                    "library:ci-fiat_crypto_legacy"
+                then "\nping @JasonGross"
+                else "" )
+            in
+            match bot_info.github_token with
+            | ACCESS_TOKEN _t ->
+                Lwt.return ()
```
Zimmi48, line 50: Add a comment to say that we report allow failure errors only when we have access to the Checks API.

src/actions.ml
```diff
@@ -28,28 +33,39 @@ let send_status_check ~bot_info job_info pr_num (gh_owner, gh_repo)
...
+                    GitHub_mutations.create_check_run ~bot_info ~name:context
+                      ~repo_id ~head_sha:job_info.commit ~conclusion:NEUTRAL
+                      ~status:COMPLETED
+                      ~title:(failure_reason ^ " on GitLab CI")
+                      ~text:"" ~details_url:job_url ~summary:message
+                    <&>
+                    if
+                      String.equal job_info.build_name
+                        "library:ci-fiat_crypto_legacy"
+                    then GitHub_mutations.post_comment ~bot_info ~id ~message
```
Zimmi48, line 65: Sorry, I'm contradicting myself, but I've come to think that since we still want to post a comment in this case, then (in this case), we should still check whether we are on the latest commit on the PR to avoid spamming when a job has failed simply because the PR has been force-pushed (which regularly happens).

Zimmi48 (commented):
Sorry still something to change (my fault).

Zimmi48 (approved):

PR #100, opened 2020-08-26
Check if a repository has the app installed.

jtcoolen:
Depends on #93. The last commit brings the desired functionality.

### DIFF
--- a/bot-components/GitHub_app.ml
+++ b/bot-components/GitHub_app.ml
@@ -89,3 +89,23 @@ let get_installation_token ~bot_info ~key ~app_id ~owner ~repo =
       get_installation_token ~bot_info ~jwt ~owner ~repo
   | Error e ->
       Lwt.return (Error e)
+
+let get_installations ~bot_info ~key ~app_id =
+  match make_jwt ~key ~app_id with
+  | Ok jwt -> (
+      get ~bot_info ~token:jwt ~url:"https://api.github.com/app/installations"
+      >|= fun body ->
+      try
+        let json = Yojson.Basic.from_string body in
+        let open Yojson.Basic.Util in
+        Ok
+          ( json |> to_list
+          |> List.map ~f:(fun json ->
+                 json |> member "account" |> member "login" |> to_string) )
+      with
+      | Yojson.Json_error err ->
+          Error (f "Json error: %s" err)
+      | Yojson.Basic.Util.Type_error (err, _) ->
+          Error (f "Json type error: %s" err) )
+  | Error e ->
+      Lwt.return (Error e)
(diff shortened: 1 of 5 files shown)


src/github_installations.ml
```diff
@@ -42,3 +39,21 @@ let action_as_github_app ~bot_info ~key ~app_id ~owner ~repo action =
...
         action
+
+let action_as_github_app ~bot_info ~key ~app_id ~owner ~repo action =
+  (* Executes an action with an installation token if the repository has
+     the GitHub app installed.
+     Generates a new installation token if the existing one has expired. *)
+  match GitHub_app.make_jwt ~key ~app_id with
+  | Ok jwt -> (
+      GitHub_queries.get_installations
+        ~bot_info:{bot_info with github_token= INSTALL_TOKEN jwt}
```
Zimmi48, line 50: If you put `get_installations` inside `GitHub_app` instead of `GitHub_queries`, you wouldn't have to export the `make_jwt` function and abuse the `INSTALL_TOKEN` constructor to store something that is not an install token.

Furthermore, it seems to me that there is a problem with the order of the calls. Why do a query to GitHub before even checking if we have a token in store for this owner? This is going to add many unnecessary requests.

Zimmi48:
In theory (though I can't figure out how to subscribe to it), there exists an [`installation` webhook](https://developer.github.com/webhooks/event-payloads/#installation) for GitHub Apps. Ideally, we could rely on it in the following way:
- at initialization time, we use the GitHub query to learn about all existing installations, we store those in a hash table,
- we use the events sent by this webhook to learn about any modifications (in particular new installations) and update the hash table accordingly.
As a result, we never have to query GitHub again to learn about the installations.

bot-components/GitHub_app.ml
```diff
@@ -89,3 +89,15 @@ let get_installation_token ~bot_info ~key ~app_id ~owner ~repo =
       get_installation_token ~bot_info ~jwt ~owner ~repo
   | Error e ->
       Lwt.return (Error e)
+
+let get_installations ~bot_info ~key ~app_id =
+  match make_jwt ~key ~app_id with
+  | Ok jwt ->
+      generic_get ~bot_info:{bot_info with github_token= JWT jwt}
```
Zimmi48, line 96: Why use the `generic_get` function here (at the cost of having to add the new `JWT` constructor to the `github_token` type)? Doesn't the `get` function that it is defined in this file work?

Zimmi48 (approved):
LGTM now. Let's wait for the base PR to be merged to rebase this one and merge it afterwards (with just the last three commits).

PR #112, opened 2020-10-07
Minimize failed tests.

### LINKED ISSUE #107: coqbot should be able to automatically turn CI failures into test-suite examples
Following up on https://github.com/coq/coq/pull/7825#issuecomment-689666091 , I think it'd be really great if, on any PR where the test-suite succeeded but some CI failed (and that CI succeeded on master), coqbot spun up the bug-minimizer and attempted to minimize the CI failure into something that could be added to the test-suite.

This will take a bit of extra work from my side, since I want to have the minimizer check two versions of Coq (and ensure that one is succeeding and one is failing; it's no use, e.g., minimizing a typeclass failure into a smaller one if the smaller one fails on both master and the branch).

There's a couple ways we could do this:
1. coqbot could pull the commit hashes of current master and the commit that failed, get the URL of the Coq repo holding the PR commit, generate a script for building the CI development and all of its dependencies (lifted from the CI scripts), and feed this information to the minimizer
2. coqbot could pull the commit hashes and URL, and the name of the makefile target that will build the CI development (such as `make ci-rewriter` or whatever) and have the minimizer repo just extract what it needs from there
3. coqbot could just point the minimizer at the PR number and ask the minimizer to do the rest

While 3 would be easiest for coqbot, it makes it much harder to coordinate about which CI needs to be minimized, and runs the risk of the PR branch being out of date by the time the minimizer runs.  Therefore I'd prefer 2, or even 1 if it's not too much work.

Is this doable?

Zimmi48:
When a PR pipeline completes, looks for the library builds that failed in the PR but did not fail for the base commit.
For now, nothing is done besides adding to the log.

When this is finished, this will resolve #107.

### DIFF
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -799,15 +799,6 @@ let suggest_ci_minimization_for_pr = function
 
 let minimize_failed_tests ~bot_info ~owner ~repo ~pr_number ~base ~head
     ~head_pipeline_summary ~request =
-  let pluralize word ?(plural = None) ls =
-    match (ls, plural) with
-    | [_], None ->
-        word ^ "s"
-    | [_], Some plural ->
-        plural
-    | _, _ ->
-        word
-  in
   fetch_ci_minimization_info ~bot_info ~owner ~repo ~pr_number ~base ~head
     ~head_pipeline_summary
   >>= function
@@ -869,6 +860,15 @@ let minimize_failed_tests ~bot_info ~owner ~repo ~pr_number ~base ~head
       run_ci_minimization ~bot_info ~comment_thread_id ~owner ~repo ~base ~head
         ~ci_minimization_infos:jobs_to_minimize
       >>= fun (jobs_minimized, jobs_that_could_not_be_minimized) ->
+      let pluralize word ?(plural = None) ls =
+        match (ls, plural) with
+        | [_], None ->
+            word ^ "s"
+        | [_], Some plural ->
+            plural
+        | _, _ ->
+            word
+      in
       (* Construct a comment body *)
       let unminimizable_jobs_description ~f =
         match
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -676,6 +676,8 @@ let minimize_failed_tests ~bot_info ~owner ~repo ~pr_number ~base ~head
                         Lwt_io.printf "Ignoring these failed tests: %s"
                           (String.concat ~sep:", " ignored_tests) )
                     >>= fun () ->
+                    init_git_bare_repository ~bot_info
+                    >>= fun () ->
                     to_minimize
                     |> Lwt_list.iter_p
                          (minimize_failed_test ~bot_info ~owner ~repo ~pr_id
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -145,7 +145,7 @@ let send_status_check ~bot_info job_info ~pr_num (gh_owner, gh_repo)
             >>= function
             | Ok repo_id ->
                 GitHub_mutations.create_check_run ~bot_info ~name:context
-                  ~repo_id ~head_sha:job_info.common_info.commit
+                  ~repo_id ~head_sha:job_info.common_info.head_commit
                   ~conclusion:NEUTRAL ~status:COMPLETED ~title
                   ~details_url:job_url
                   ~summary:("This job is allowed to fail.\n\n" ^ summary_tail)
@@ -166,15 +166,15 @@ let send_status_check ~bot_info job_info ~pr_num (gh_owner, gh_repo)
           >>= function
           | Ok {issue= id; head}
           (* Commits reported back by get_pull_request_refs are surrounded in double quotes *)
-            when String.equal head.sha (f "\"%s\"" job_info.common_info.commit)
-            ->
+            when String.equal head.sha
+                   (f "\"%s\"" job_info.common_info.head_commit) ->
               GitHub_mutations.post_comment ~bot_info ~id ~message
               >>= GitHub_mutations.report_on_posting_comment
           | Ok {head} ->
               Lwt_io.printf
                 "We are on a PR branch but the commit (%s) is not the current \
                  head of the PR (%s). Doing nothing.\n"
-                job_info.common_info.commit head.sha
+                job_info.common_info.head_commit head.sha
           | Error err ->
               Lwt_io.printf
                 "Couldn't get a database id for %s#%d because the following \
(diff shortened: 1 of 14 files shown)


src/actions.ml
```diff
@@ -327,22 +306,61 @@ let job_action ~bot_info (job_info : job_info) ~gitlab_mapping =
...
+    library_name base head
+
+let minimize_failed_tests ~bot_info ~owner ~repo ~base ~head =
+  GitHub_queries.get_base_and_head_checks ~bot_info ~owner ~repo ~base ~head
+  >>= function
+  | Error err ->
+      Lwt_io.printf
+        "Error while looking for failed library tests to minimize: %s\n" err
+  | Ok (base_checks, head_checks) ->
+      (* TODO: do not run any test if test-suite is updated. *)
```
JasonGross, line 341: I think the current behavior is correct and this TODO is not needed: if the test-suite is updated and fails, then we already don't try to minimize.  If the test-suite is updated and succeeds, but the CI still fails, then this means there's a test-case in the CI which is not tracked by the test-suite, right?

Zimmi48:
Still remaining to-do: the report after we get back the results.

src/actions.ml
```diff
@@ -538,40 +777,53 @@ let pipeline_action ~bot_info pipeline_info ~gitlab_mapping : unit Lwt.t =
...
+              Lwt_unix.sleep 5.
+              >>= fun () ->
+              match
+                ( owner
+                , repo
+                , pipeline_info.state
+                , pipeline_info.common_info.base_commit )
+              with
+              | "coq", "coq", "failed", Some base_commit ->
+                  minimize_failed_tests ~bot_info ~owner ~repo ~base:base_commit
```
JasonGross, line 803: Idk if this is the right place for it, but
```suggestion
                  init_git_bare_repository ~bot_info
                  >>= fun () ->
                  minimize_failed_tests ~bot_info ~owner ~repo ~base:base_commit
```

Zimmi48:
From the logs, it looks like we check the argument count (and forgot to update it) but this doesn't prevent running the rest of the script!

```
22 Apr 2021 18:40:33.944691 <190>1 2021-04-22T16:40:33.455235+00:00 app web.1 - - Executing command: ./run_ci_minimization.sh 'MDExOlB1bGxSZXF1ZXN0NjE5MTcxMDA5' 'ad6b0bc018138064e6f5e313f6a645e2519755ac' 'coqbot' 'coqbot.herokuapp.com' 'coq' 'coq' 'registry.gitlab.com/coq/coq:bionic_coq-V2021-04-14-802aebab96' 'ci-metacoq' '4.05.0' 'https://gitlab.com/coq/coq/-/jobs/1194985509/artifacts/download https://gitlab.com/coq/coq/-/jobs/1203588043/artifacts/download' 'https://gitlab.com/coq/coq/-/jobs/1194892258/artifacts/download https://gitlab.com/coq/coq/-/jobs/1194892386/artifacts/download' 'cab7a5ddb906e5cef57d78ba7435e89354f3125b' '6a8e17d6344d9189fdff571db9967787887681c6'
22 Apr 2021 18:40:33.944113 <190>1 2021-04-22T16:40:33.457479+00:00 app web.1 - - + '[' 13 '!=' 8 ']'
22 Apr 2021 18:40:33.944119 <190>1 2021-04-22T16:40:33.457481+00:00 app web.1 - - + echo Bad argument count
22 Apr 2021 18:40:33.944112 <190>1 2021-04-22T16:40:33.457501+00:00 app web.1 - - Bad argument count
```

Zimmi48:
Another bug is that it triggered minimization on a commit which also had test-suite failures:
https://github.com/coq/coq/pull/14153#issuecomment-825065261

ejgallego:
Hi folks, some feedback from discuss:
- it could be good to make this opt-in first `@coqbot minimize library:foo` , and once the setup is better tested add some automatic triggering
- rules for automation are a bit complex, but some PRs such as `kind: infrastructure` should likely no be ran automatically, whether to use tags to forbid or enable minimization is not clear.
- logic for automatic running is a bit complex, for example what about overlays? Not sure I got some ideas here but I will think.

JasonGross:
> logic for automatic running is a bit complex, for example what about overlays? Not sure I got some ideas here but I will think.

I guess there is a presumption that overlay authors will check the overlay against either the base commit or the PR.  If this is not the case, then indeed it perhaps makes sense to exclude overlays (or to have coqbot comment with a reminder that overlays can be triggered separately).

Here is one proposal:
- if the user says "coqbot ci minimize", then we run the failures that we can find
- if the user says "coqbot ci overlays minimize", we run only the failed checks which are overlayed
- if the user says "coqbot minimize ci-...", we minimize only that one
- if it is possible to minimize tests, then we inform the user that they can trigger minimization
- once things are working well, we can perhaps automatically trigger minimization on non-infrastructure PRs and comment about the fact that minimization is available/happening

Zimmi48:
Yet another bug is `Couldn't find base build job url for build:edge+flambda in:` and then the text printed contains ` - [build:edge+flambda](https://gitlab.com/coq/coq/-/jobs/1202903267)`.

Yet another surprising behavior (though in this case the result is the one we would like): `Ignored failed test GitLab CI job plugin:plugin-tutorial (pull request) (not a library nor a plugin).`

We might also want to avoid running all the CI minimization scripts in parallel as this create more risk of problems. I've seen several occurrences of:
```
22 Apr 2021 22:27:14.593137 <190>1 2021-04-22T20:27:14.345352+00:00 app web.1 - - fatal: bad object worktrees/tmp.NLccOP/HEAD
22 Apr 2021 22:27:14.593194 <190>1 2021-04-22T20:27:14.345625+00:00 app web.1 - - error: https://github.com/coq-community/run-coq-bug-minimizer.git did not send all necessary objects
```
when there were many failures in the same PR.

Slightly unrelated to the actual minimization feature, I've also noticed that sometimes for errors in OCaml, the filename is not on the line just above the error message: this was the case for https://github.com/coq/coq/pull/6285/checks?check_run_id=2413105482.

JasonGross:
> Slightly unrelated to the actual minimization feature, I've also noticed that sometimes for errors in OCaml, the filename is not on the line just above the error message: this was the case for https://github.com/coq/coq/pull/6285/checks?check_run_id=2413105482.

Ah, yes, I guess we want to search backwards from the "Error" looking for "File"?

JasonGross:
I've finished my refactoring at https://github.com/Zimmi48/bot/pull/1

Zimmi48:
@JasonGross I've merged #140, thus the current PR is not deployed anymore. Please rebase on top of `master` before redeploying.

BTW, when you feel this is ready, maybe there will be a bit of history-rewriting to do? I am not super strict on commit history in this repository (no problem if you want to keep many commits), but commit messages like "WIP" do not feel like they should get merged as is.

JasonGross:
Rebased.  I can do some history rewriting when I get a chance

PR #144, opened 2021-06-09
Fix quoting and regex parsing for minimization

### LINKED ISSUE #134: coqbot minimize should have proper quoting of the script
Currently the regex at 
https://github.com/coq/bot/blob/b10cfb9dd8aea4eb5865f3d636123f1794312eca/src/bot.ml#L160
will be incorrect if the script contains a backtick (``` ` ```) (it should just look for `@coqbot:? [Mm]inimize` followed by a newline, followed by a line starting with ` ``` `, and then take the script to be all of the lines until the next line starting with ` ``` `).  Additionally, the invocation of the bash script at
https://github.com/coq/bot/blob/b10cfb9dd8aea4eb5865f3d636123f1794312eca/src/git_utils.ml#L120-L121
just sticks the script between single quotes (`'`), which is not proper quoting.  Instead, [`Filename.quote_command`](https://ocaml.org/api/Filename.html#VALquote_command) should be used.

JasonGross:
On top of #112 

Fixes #134

### DIFF
--- a/src/bot.ml
+++ b/src/bot.ml
@@ -45,6 +45,19 @@ let string_of_installation_tokens =
 
 let callback _conn req body =
   let body = Cohttp_lwt.Body.to_string body in
+  let coqbot_minimize_text_of_body body =
+    let body = Helpers.trim_comments body in
+    if
+      string_match
+        ~regexp:(f "@%s:? [Mm]inimize[^`]*```[^\n]*\n\\(.+\\)" bot_name)
+        body
+    then
+      Some
+        ( Str.matched_group 1 body
+        |> Str.split (Str.regexp_string "```")
+        |> List.hd |> Option.value ~default:"" )
+    else None
+  in
   (* print_endline "Request received."; *)
   match Uri.path (Request.uri req) with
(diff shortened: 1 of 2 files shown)


Zimmi48 (approved):

src/bot.ml
```diff
@@ -153,98 +166,94 @@ let callback _conn req body =
...
                 ~owner:issue_info.issue.owner ~repo:issue_info.issue.repo
-                (run_coq_minimizer ~script:(Str.matched_group 1 body)
-                   ~comment_thread_id:issue_info.id
+                (run_coq_minimizer ~script ~comment_thread_id:issue_info.id
                    ~comment_author:issue_info.user ~owner:issue_info.issue.owner
                    ~repo:issue_info.issue.repo))
             |> Lwt.async ;
-          Server.respond_string ~status:`OK ~body:"Handling minimization." ()
-      | Ok (signed, CommentCreated comment_info) ->
-          let body = Helpers.trim_comments comment_info.body in
```
JasonGross, line 175: Oops, I should not have removed this

PR #146, opened 2021-06-09
Fix a parsing bug I introduced in https://github.com/coq/bot/pull/144/files#r648423852

JasonGross:
Fix a parsing bug I introduced in https://github.com/coq/bot/pull/144/files#r648423852

I wonder if we should do less shadowing...

### DIFF
--- a/src/bot.ml
+++ b/src/bot.ml
@@ -46,7 +46,6 @@ let string_of_installation_tokens =
 let callback _conn req body =
   let body = Cohttp_lwt.Body.to_string body in
   let coqbot_minimize_text_of_body body =
-    let body = Helpers.trim_comments body in
     if
       string_match
         ~regexp:(f "@%s:? [Mm]inimize[^`]*```[^\n]*\n\\(.+\\)" bot_name)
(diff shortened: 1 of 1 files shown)


src/bot.ml
```diff
@@ -167,93 +166,97 @@ let callback _conn req body =
...
-                  ~owner:comment_info.issue.issue.owner
-                  ~repo:comment_info.issue.issue.repo
-                  (ci_minimize ~comment_info ~requests ~comment_on_error:true))
+                  ~owner:issue_info.issue.owner ~repo:issue_info.issue.repo
+                  (run_coq_minimizer ~script ~comment_thread_id:issue_info.id
+                     ~comment_author:issue_info.user
+                     ~owner:issue_info.issue.owner ~repo:issue_info.issue.repo))
               |> Lwt.async ;
+              Server.respond_string ~status:`OK ~body:"Handling minimization."
+                ()
```
JasonGross, line 182: @Zimmi48 Do you understand why ocamlformat introduced this change?  Did I just forget to run ocamlformat on a previous commit, or does it sometimes make non-local changes like this?

Zimmi48: Or a bug.

Zimmi48: I just tried it out. I have ocamlformat 0.15.0. Running `dune build @fmt` on the current `master` succeeds (with no changes). Same when running on the PR.

Zimmi48: I tried it out with `ocamlformat` 0.18 and (despite other formatting changes), the problem is the same. It might be worth a bug report to the `ocamlformat` people, but I'm not sure whether it's worth the effort.

Zimmi48 (approved):
LGTM. I'll let you merge this and the other PR in the correct order. After that, I'll take over on the coqbot code.

JasonGross:
Thanks!  Now merged

PR #148, opened 2021-06-17
Enable CI minimization resumption

### LINKED ISSUE #143: coqbot should be able to resume CI minimization
I'd like to be able to write things like

    @coqbot resume ci minimization ci-hott with
    ```
    some coq file here
    ```

And have minimization resume on the given file.  This would be useful at, for example, https://github.com/coq/coq/pull/13269#issuecomment-854117817, where minimization times out.  Most of the work to be done here is on my side on https://github.com/coq-community/run-coq-bug-minimizer; the only thing to do on the coqbot side is to recognize this sort of request and initiate CI minimization with an extra file of an appropriate name for the minimizer runner to recognize it.

JasonGross:
Fixes #143

### DIFF
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -472,14 +472,25 @@ type ci_minimization_info =
   ; passing_urls: string }
 
 let run_ci_minimization ~bot_info ~comment_thread_id ~owner ~repo ~base ~head
-    ~ci_minimization_infos =
-  Lwt_list.map_s
-    (fun {target; opam_switch; failing_urls; passing_urls; docker_image} ->
-      git_run_ci_minimization ~bot_info ~comment_thread_id ~owner ~repo
-        ~docker_image ~target ~opam_switch ~failing_urls ~passing_urls ~base
-        ~head
-      >>= fun result -> Lwt.return (target, result))
-    ci_minimization_infos
+    ~ci_minimization_infos ~bug_file_contents =
+  (* for convenience of control flow, we always create the temporary
+     file, but we only pass in the file name if the bug file contents
+     is non-None *)
+  Lwt_io.with_temp_file (fun (bug_file_name, bug_file_ch) ->
+      Lwt_io.write bug_file_ch (Option.value ~default:"" bug_file_contents)
+      >>= fun () ->
+      Lwt_io.flush bug_file_ch
+      >>= fun () ->
+      let bug_file_name =
+        Option.map ~f:(fun _ -> bug_file_name) bug_file_contents
+      in
+      Lwt_list.map_s
+        (fun {target; opam_switch; failing_urls; passing_urls; docker_image} ->
+          git_run_ci_minimization ~bot_info ~comment_thread_id ~owner ~repo
+            ~docker_image ~target ~opam_switch ~failing_urls ~passing_urls ~base
+            ~head ~bug_file_name
+          >>= fun result -> Lwt.return (target, result))
+        ci_minimization_infos)
   >>= fun results ->
   results
   |> List.partition_map ~f:(function
@@ -879,7 +890,7 @@ let suggest_ci_minimization_for_pr = function
       Suggest
 
 let minimize_failed_tests ~bot_info ~owner ~repo ~pr_number
-    ~head_pipeline_summary ~request ~comment_on_error =
+    ~head_pipeline_summary ~request ~comment_on_error ~bug_file_contents =
   fetch_ci_minimization_info ~bot_info ~owner ~repo ~pr_number
     ~head_pipeline_summary
   >>= function
@@ -964,7 +975,7 @@ let minimize_failed_tests ~bot_info ~owner ~repo ~pr_number
             |> String.concat ~sep:", " ) )
       >>= fun () ->
       run_ci_minimization ~bot_info ~comment_thread_id ~owner ~repo ~base ~head
-        ~ci_minimization_infos:jobs_to_minimize
+        ~ci_minimization_infos:jobs_to_minimize ~bug_file_contents
       >>= fun (jobs_minimized, jobs_that_could_not_be_minimized) ->
       let pluralize word ?plural ls =
         match (ls, plural) with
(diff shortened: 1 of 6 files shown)


JasonGross:
I think I'm getting `17 Jun 2021 15:01:49.415141 <190>1 2021-06-17T19:01:49.001787+00:00 app web.1 - - Cohttp_lwt: body not consumed - leaking stream!` from this; is this an issue @Zimmi48 ?

Zimmi48:
Not an issue AFAICT. It was already happening before and I didn't yet take care of fixing it.

JasonGross:
Seems to be working well at https://github.com/coq-community/run-coq-bug-minimizer/runs/2854121247?check_suite_focus=true / https://github.com/coq/coq/pull/14328#issuecomment-863490227

src/actions.ml
```diff
@@ -1550,6 +1563,77 @@ let coq_bug_minimizer_results_action ~bot_info ~ci ~key ~app_id body =
...
+          :: target
+             :: opam_switch
+                :: failing_urls
+                   :: passing_urls :: base :: head :: bug_file_lines ->
+            (let bug_file_contents = String.concat ~sep:"\n" bug_file_lines in
+             fun () ->
+               init_git_bare_repository ~bot_info
+               >>= fun () ->
+               Github_installations.action_as_github_app ~bot_info ~key ~app_id
+                 ~owner ~repo
```
Zimmi48, line 1585: I'm a bit surprised to see this here instead of in the call to the action function from `bot.ml`.

Zimmi48 (commented):
IIUC the new endpoint is triggered by the GitHub Action directly. But you also added a `@coqbot: resume minimization` feature. Have you tested this one as well?

JasonGross:
> Have you tested this one as well?

I have not, let me go test that

src/actions.ml
JasonGross: I was copying https://github.com/coq/bot/blob/e7327b50da0d8aeaa68cd4ebb137cef0224aa233/src/actions.ml#L1523-L1534
Should both of these be moved to bot.ml?

JasonGross:
Good call, explicit resumption requests seem to result in an empty bug.v file:
https://github.com/coq/coq/pull/14328#issuecomment-864129759
https://github.com/coq-community/run-coq-bug-minimizer/commit/29a604e4d5c56e1dfb3965a6f97698969c8217d1#diff-651444b27c992ddb257e297752de3755d12fcb603effec3370687b46d5ae0616

src/actions.ml
Zimmi48: Probably yes.

Zimmi48:
Two more questions:
- Can it happen that the minimizer times out without having succeeded to minimize the file further? In this case, is there a risk of getting into an infinite loop?
- What is the use case for explicitly requesting to resume minimization if the bot already takes care of triggering this?

JasonGross:
> Can it happen that the minimizer times out without having succeeded to minimize the file further? In this case, is there a risk of getting into an infinite loop?

This will only happen if the initial Coq run of the single file takes more than ~1.5 hours.  The minimizer will refuse to make changes that drastically increase the running time (other than inlining files).  So indeed an infinite loop is possible, but we'll notice it because we'll get the same comment again and again on the issue, and I believe we can break the infinite loop just by canceling any of the GH actions runs.  It should be pretty rare, but if it starts showing up, I can add a couple of lines to disable resumption if the initial and final file are identical modulo the comments on the first couple lines.

> What is the use case for explicitly requesting to resume minimization if the bot already takes care of triggering this?

If I want to make some edits to the file locally (to do some minimization that the minimizer can't do) and then resume minimization against the version of Coq on the PR, this is the way to do it.

src/actions.ml
JasonGross: Oh, the issue is that we don't have access to owner and repo from `bot.ml`; they are only accessible after we parse the stamp

JasonGross:
> Good call, explicit resumption requests seem to result in an empty bug.v file:
> [coq/coq#14328 (comment)](https://github.com/coq/coq/pull/14328#issuecomment-864129759)
> [coq-community/run-coq-bug-minimizer@29a604e#diff-651444b27c992ddb257e297752de3755d12fcb603effec3370687b46d5ae0616](https://github.com/coq-community/run-coq-bug-minimizer/commit/29a604e4d5c56e1dfb3965a6f97698969c8217d1#diff-651444b27c992ddb257e297752de3755d12fcb603effec3370687b46d5ae0616)

Updated, now seems to work at https://github.com/coq-community/run-coq-bug-minimizer/commit/3c1bc33de2bf57b68bf157f1731ad4661d60d8f8#diff-651444b27c992ddb257e297752de3755d12fcb603effec3370687b46d5ae0616 and https://github.com/coq/coq/pull/14328#issuecomment-864158013

src/actions.ml
Zimmi48: Oh I see.

Zimmi48 (approved):
Feel free to merge whenever you feel like.

JasonGross:
Here's an example of me making good use of manual minimization resumption: https://github.com/coq/coq/pull/14328#issuecomment-864594582
I did some local minimization to work around https://github.com/JasonGross/coq-tools/issues/67, and then let the minimizer run locally to the point where the file was small enough to paste into a comment, and then asked coqbot to continue minimizing it on GH Actions.

Also, I did run into a different infinite loop case: when a file fails to be inlined, if the minimizer does not find any inlinable file within the time limit, this can result in an infinite loop, because the blacklist of which files to skip the inlining of is not preserved across separate runs of the bug minimizer.  I've improved the computation of the blacklist to recursively blacklist dependent files when any single file fails to inline, which should help with this infinite loop (and it did, in practice).  But I think that infinite loops are not so bad, as we can always just cancel the job.

Zimmi48:
>But I think that infinite loops are not so bad, as we can always just cancel the job.

Yes and no. One needs to be a coq-community member to stop the pipeline.

JasonGross:
Hm, well, I could add support for persisting the blacklist across calls.  Perhaps even just add a line to the standard file comment header itself, and then no modifications are needed to the bot.  I suspect this case will be fairly rare, though, and I'm inclined to wait until it actually starts coming up in practice.  (It's not too hard to tag me or open an issue on one of the repos about a runaway loop.)

PR #149, opened 2021-06-20
Fix coqbot minimize regex

JasonGross:
According to https://ocaml.org/api/Str.html, `.` "[m]atches any
character except newline."  It was previously only catching the first line.  Oops.

### DIFF
--- a/src/bot.ml
+++ b/src/bot.ml
@@ -53,7 +53,8 @@ let callback _conn req body =
   let coqbot_minimize_text_of_body body =
     if
       string_match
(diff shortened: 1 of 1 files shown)


JasonGross:
I guess I should have tested it before with multiple lines.  This has now been tested at https://github.com/coq/coq/pull/14328#issuecomment-864602242 / https://github.com/coq-community/run-coq-bug-minimizer/commit/a3f0578a3d4a515ec5793956d903971725a2c823#diff-651444b27c992ddb257e297752de3755d12fcb603effec3370687b46d5ae0616 / https://github.com/coq-community/run-coq-bug-minimizer/runs/2870234329 and seems to be working

Zimmi48 (approved):

PR #152, opened 2021-06-22
Better `coqbot ci minimize all` messages

JasonGross:


JasonGross:
@Zimmi48 Is this mergable?

JasonGross:
And apparently there was a bug in the old version of this PR (where I accidentally replaced `comment_info.body` with `body` in much the same way that I made the mistake that was fixed by #146).  Now fixed.

src/bot.ml
```diff
@@ -59,6 +59,15 @@ let callback _conn req body =
...
     else None
   in
+  let strip_quoted_bot_name body =
+    (* If someone says "`@coqbot minimize foo`", (with backticks), we
+       don't want to treat that as them tagging coqbot, so we adjust
+       the tagging to "@`coqbot minimize foo`" so that the matching
+       below doesn't pick up the name *)
+    Str.global_replace
+      (Str.regexp_string (f "`@%s " bot_name))
+      (f "@`%s " bot_name) body
```
Zimmi48, line 69: Ideally we would go further and come up with a parser for coqbot request that doesn't look inside quoted code. But well, this is good enough for now.

Zimmi48 (approved):

PR #153, opened 2021-06-25
Introduce a new hook to check for stale PRs.

ppedrot:
This adds a `check-stale-pr` message to the possible actions of coqbot. This hook checks for PRs that have the "needs: rebase" label, warns if it has been there for >= 30 days and closes it if it's been there for >= 60 days. It accepts a payload of the form organization:repo to be acted upon.

I have not tested it yet, but I am a bit reluctant to install the docker *usine à gaz* on my machine just for that. I am not even sure this is the correct way to test it for the official coq repo.

Things to do in parallel of this PR:
- ~~Refresh all the "needs: rebase" tags on the Coq repo not to let the bot go postal and close about half of the currently open PRs. At least our poor devs deserve a fair warning.~~ We throttle instead.
- Add the relevant workflow to coq/coq. A simple curl invocation should be enough.


Zimmi48 (commented):
>I have not tested it yet, but I am a bit reluctant to install the docker usine à gaz on my machine just for that. I am not even sure this is the correct way to test it for the official coq repo.

In this case, it would probably work well, but it's not required. We most often test by deploying. Because of the way the bot is architectured, there isn't much risk in doing that (apart of course if there is a business logic mistake that will close all the PRs, but this could also happen while testing locally).

>Refresh all the "needs: rebase" tags on the Coq repo not to let the bot go postal and close about half of the currently open PRs. At least our poor devs deserve a fair warning.

What about adding a new `stale` label signalling that the warning was issued? Then, you can count the number of days since the `stale` label was added instead of the number of days since the `needs: rebase` label was added to decide when to close. Of course, if we do this, then we need to adapt the `needs: rebase` removal logic to also remove the `stale` label.

Furthermore, to avoid notification hell on the day when this feature is deployed, you could cap the number of warnings that the bot would deliver (e.g. to 5 or 10).

bot-components/GitHub_queries.ml
```diff
@@ -665,3 +665,72 @@ let get_cards_in_column column_id =
...
+    | Error err ->
+      Lwt.return @@ Error err
+  in
+  eval None []
+
+let get_pull_request_label_timeline ~bot_info ~owner ~repo ~pr_number =
+  let rec eval cursor accu =
+    GitHub_GraphQL.GetPullRequestLabelTimeline.make ~owner ~repo ~prNumber:pr_number ~len:100 ?cursor ()
+    |> GraphQL_query.send_graphql_query ~bot_info
+    >>= function
```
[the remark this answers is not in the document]
ppedrot: I've added a combinator but it looks worse to me now.

bot-components/GitHub_queries.ml
```diff
@@ -665,3 +665,72 @@ let get_cards_in_column column_id =
...
+        | None -> []
+        | Some data ->
+          let map = function None -> None | Some o -> Some (o#id, o#number) in
+          List.filter_map ~f:map (Array.to_list data)
+        in
+        let accu = List.rev_append data accu in
+        match next with
+        | None -> Lwt.return (Ok (List.rev accu))
+        | Some cursor -> eval (Some cursor) accu
+        )
```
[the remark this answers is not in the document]
ppedrot: From my interaction with the GraphQL API it seemed to be the case though. When there are no more pages, it returns `null`. It's not specified so it's hard to tell whether this is supposed to be always the case.

ppedrot:
> What about adding a new stale label signalling that the warning was issued?

This approach has both advantages and inconvenients. The advantage is that we decorrelate staleness from the mere status of rebase of the PR, which allows to implement finer control. It is also better for the transient situation we're in, namely having a lot of PRs that are morally stale but were not explicitly marked as such. The inconvenient is that it requires to make explicit the mechanism by which we consider that a PR is not stale anymore, which is moving the complexity on the dual side of the "needs: rebase" label. For now a PR is stale iff it had a given label for more than X days, but if we make this more complex, we'll have to think a bit what it means to not be stale anymore. Is is a conjunction of the introducing conditions? Something else?

Zimmi48:
The `stale` label was just an idea to easily retrieve when the warning was issued. Another solution is to just look for the coqbot comment in the timeline. In this case, given that a PR can have way more comments than label events, we would need to retrieve the timeline lazily, from bottom to top, counting on the fact that what we are looking for (`needs: rebase` label or coqbot comment) is probably going to be close to the bottom, so the operation won't be too costly.

ppedrot:
I actually quite like the stale label solution, despite the added complexity. I'll write a patch atop of this one to implement it.

ppedrot:
This should work using the stale label now. I have not implemented the throttling issue yet, though. I acknowledge it will be a bit annoying to be drowned under 100+ label notifications when the bot is set free, but it's also annoying that we have to tweak the code for a single occasion and weaken it voluntarily. This throttling wouldn't be an issue once we get to a stationary regime.

Zimmi48:
What about the capping idea? This would be useful not only in the beginning, but also in the future, to avoid many notifications one month after a developer rebased all of their PRs.

ppedrot:
> avoid many notifications one month after a developer rebased all of their PRs

My secret masterplan does include the fact that bot should be annoying so that people feel a stronger incentive to close PRs that are lingering without any reason. Trying to be nice to devs that have many bitrotten PRs kinds of defeats the purpose...

(I am going to implement the throttling though, since you insist so much.)

ejgallego:
> My secret masterplan does include the fact that bot should be annoying so that people feel a stronger incentive to close PRs that are lingering without any reason.

I'd bet this will work the other way, and the annoyed developers will just ask for the bot to be obliterated.

Zimmi48:
> I'd bet this will work the other way, and the annoyed developers will just ask for the bot to be obliterated.

Indeed, that's how it works in my experience :wink:

ppedrot:
After the last commit coqbot shouldn't mark more than 10 PRs stale per call to the bot.

ppedrot:
So the two remaining points to solve before merging are:
- Should I add the ISO8601 package to the opam dependencies despite being transitively required?
- Should I rely on `hasNextPage` to paginate results, despite the API seemingly returning a `null` cursor when there are no more results?

Zimmi48:
> Should I add the ISO8601 package to the opam dependencies despite being transitively required?

I think so.

>Should I rely on `hasNextPage` to paginate results, despite the API seemingly returning a `null` cursor when there are no more results?

Should this behavior change in the future, it would create an infinite loop, so this would be more robust indeed.

ppedrot:
Should be ready now.

ppedrot:
@Zimmi48 would you assign?

bot-components/GitHub_mutations.ml
```diff
@@ -141,30 +150,17 @@ let update_check_run ~bot_info ~check_run_id ~repo_id ~conclusion ?details_url
 
 (* TODO: use GraphQL API *)
```
Zimmi48, line 151: This comment will need to move below since the two functions below were converted to use GraphQL.

src/actions.ml
```diff
@@ -1795,12 +1795,23 @@ let update_pr ~bot_info (pr_info : issue_info pull_request_info) ~gitlab_mapping
           ~pr_number:pr_info.issue.number ~base:local_base_branch
           local_head_branch)
   >>= fun ok ->
+  let label = "needs: rebase" in
+  let issue = pr_info.issue.issue in
+  GitHub_queries.get_label ~bot_info ~owner:issue.owner ~repo:issue.repo ~label >>= fun label_id ->
+  GitHub_queries.get_label ~bot_info ~owner:issue.owner ~repo:issue.repo ~label:"stale" >>= fun stale_id ->
```
Zimmi48, line 1801: Important: make it robust to these labels not existing. With the REST API call, this was fine. But now, the queries for getting the labels IDs need to succeed both to check whether one of these labels exists. At the current time (and this will probably continue being the case in the future), most repositories relying on coqbot do not have the stale label (they all have the `needs: rebase` label because the REST API call did also create the label if it didn't already exist, so this one change will need to be documented).

src/actions.ml
```diff
@@ -1795,12 +1795,23 @@ let update_pr ~bot_info (pr_info : issue_info pull_request_info) ~gitlab_mapping
...
   if ok then (
```
Zimmi48, line 1802: Note that only the query for the `needs: rebase` label really makes sense to factorize above this conditional. The one for the `stale` label could be specific to the first branch.

src/bot.ml
```diff
@@ -340,6 +340,21 @@ let callback _conn req body =
...
+      body
+      >>= fun body ->
+      (match String.split ~on:':' body with
+      | [owner; repo] ->
+        let warn_after = 30 in
+        let close_after = 30 in
+        let throttle = 10 in
+        action_as_github_app ~bot_info ~key ~app_id ~owner ~repo (coq_check_needs_rebase_pr ~owner ~repo ~warn_after ~close_after ~throttle) >>= fun () ->
+        action_as_github_app ~bot_info ~key ~app_id ~owner ~repo (coq_check_stale_pr ~owner ~repo ~after:close_after) >>= fun () ->
+        Server.respond_string ~status:`OK ~body:"Stale pull requests updated" ()
```
Zimmi48, line 353: You need to send the HTTP status back asynchronously because GitHub expects it in less than 10 seconds, which is not realistic if many other queries must be made before.

src/actions.ml
```diff
@@ -2075,3 +2086,87 @@ let push_action ~bot_info ~base_ref ~commits_msg =
...
+    end
+  | Error err ->
+    Lwt_io.print (f "Error: %s\n" err)
+
+let coq_check_needs_rebase_pr ~bot_info ~owner ~repo ~warn_after ~close_after ~throttle =
+  let label = "needs: rebase" in
+  GitHub_queries.get_label ~bot_info ~owner ~repo ~label:"stale" >>= function
+  | Ok stale_id ->
+    let action pr_id pr_number =
+      GitHub_queries.get_pull_request_labels ~bot_info ~owner ~repo ~pr_number >>= function
```
Zimmi48, line 2140: Isn't this query a bit redundant with the one that gets the label timeline in `apply_after_labels`? Wouldn't it make sense for the latter to call the action with the timeline as an extra argument?

src/actions.ml
```diff
@@ -2075,3 +2086,87 @@ let push_action ~bot_info ~base_ref ~commits_msg =
...
+    let n = List.count ~f:(fun b -> b) ans in
+    apply_throttle (len - n) action rem
+
+let apply_after_label ~bot_info ~owner ~repo ~after ~label ~action ?throttle () =
+  GitHub_queries.get_open_pull_requests_with_label ~bot_info ~owner ~repo ~label
+  >>= function
+  | Ok prs ->
+    let iter (pr_id, pr_number) =
+      GitHub_queries.get_pull_request_label_timeline ~bot_info ~owner ~repo ~pr_number >>= function
+      | Ok timeline ->
```
Zimmi48, line 2109: This isn't a change request, but we could optimize the number of API calls a lot by also returning say the last 10 timeline labeling operations within the `get_open_pull_requests_with_label` GraphQL query. Then, we would need to request the full label timeline only for the PRs for which the label was not set among the last 10 operations.

Zimmi48:
Only https://github.com/coq/bot/pull/153#discussion_r664396668 and https://github.com/coq/bot/pull/153#discussion_r664456276 actually need to be resolved before we can start testing this PR.

BTW, can you also rebase your PR on the latest `master` when you update it?

Thanks! And sorry for the delay in reviewing this.

ppedrot:
@Zimmi48 mandatory changes implemented.

Zimmi48:
OK, I'll deploy later today.

Zimmi48:
Deploy in progress: https://github.com/Zimmi48/bot/runs/2999803861?check_suite_focus=true

Zimmi48:
Tested with `curl -d "coq:coq" https://coqbot.herokuapp.com/check-stale-pr`. It worked :tada: 

See:

- https://github.com/coq/coq/pull/9122#issuecomment-874818782
- https://github.com/coq/coq/pull/7625#issuecomment-874818781
- https://github.com/coq/coq/pull/8666#issuecomment-874818784
- https://github.com/coq/coq/pull/9067#issuecomment-874818783
- https://github.com/coq/coq/pull/991#issuecomment-874818785
- https://github.com/coq/coq/pull/6836#issuecomment-874818788
- https://github.com/coq/coq/pull/7661#issuecomment-874818786
- https://github.com/coq/coq/pull/7962#issuecomment-874818791
- https://github.com/coq/coq/pull/6285#issuecomment-874818793
- https://github.com/coq/coq/pull/1003#issuecomment-874818794

ppedrot:
Well, that's half of the PR. We'll have to wait for a month now!

Zimmi48:
Right!

ejgallego:
Folks, could we maybe have the bot skip PRs that have `needs: independent fix` set?

Kind of fringe case, but I got a few PRs that do require for example a Dune update, so IMHO they could be left open as they should be merged when the time arrives.

Zimmi48:
Sure, that does make sense to me. @ppedrot do you want to take care of implementing this tweak?

ppedrot:
@Zimmi48 done.

PR #176, opened 2021-11-04
Don't suggest coq-tools for minimization

JasonGross:


src/actions.ml
```diff
@@ -507,7 +507,8 @@ type ci_minimization_job_suggestion_info =
   ; head_job_succeeded: bool
   ; missing_error: bool
   ; non_v_file: string option
-  ; job_kind: string (*; overlayed: bool*) }
+  ; job_kind: string
+  ; job_name: string (*; overlayed: bool*) }
```
Zimmi48, line 511: `job_name` might wrongly suggest that this refers to the name of the job in GitLab CI. If what is stored in there is the Makefile target, what about calling it `job_target`?

Zimmi48 (approved):

src/actions.ml
JasonGross: Done

PR #189, opened 2021-12-17
Be more lax on what merged PRs we detect, but post an alert.

Zimmi48:
It happens more often than I expected that maintainers use the merge button, even though it's explicitly forbidden in the merging rules. Examples by different maintainers:

- https://github.com/coq/coq/pull/15243
- https://github.com/coq/coq/pull/11993
- https://github.com/coq/coq/pull/11629
- https://github.com/coq/coq/pull/11605
- https://github.com/coq/coq/pull/8778

Therefore, it can happen that we miss PRs for which backporting would have been required. This is what has happened in the case of https://github.com/coq/coq/pull/11629, where the backporting process was handled manually.

This PR detects these offending PRs, but as is it is incomplete because in cases where we merge subtrees coming from other repositories, such as in the case of https://github.com/coq/coq/pull/8778, we might get confused by a bunch of merge commits that are not from the Coq repository. We can think of various ways to solve this issue.

In case we notice a wrong merge commit message, we could alert the maintainer that they are not following the documentation. Currently, all the code does is to put a warning in the log.

### DIFF
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -2108,9 +2108,24 @@ let project_action ~bot_info ~(issue : issue) ~column_id =
 let push_action ~bot_info ~base_ref ~commits_msg =
   Stdio.printf "Merge and backport commit messages:\n" ;
   let commit_action commit_msg =
-    if string_match ~regexp:"^Merge PR #\\([0-9]*\\):" commit_msg then
-      let pr_number = Str.matched_group 1 commit_msg |> Int.of_string in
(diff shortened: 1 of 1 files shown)


Zimmi48:
This more lax criterion was implemented in 22ea1a80987f7c0da6e38256744ee4ab63beb8d9, as part of #305, but without posting any alert, and without the careful considerations for the case of merging subtrees. So we might want to revisit this.

PR #205, opened 2022-03-07
Changing expiration date 

### LINKED ISSUE #204: Expiration may be too far in the future for github
https://github.com/coq/bot/blob/15a4b50d576904f5d1643ce761b85133700ecbae/bot-components/GitHub_app.ml#L21-L27 

Specifically `(issuedAt + (60 * 10))`

Since this is computed based on `Unix.time`, if your computer's date is a bit too far in the future, GitHub will complain with the following json message:
```json
{"message":"'Expiration time' claim ('exp') is too far in the future","documentation_url":"https://docs.github.com/rest"}
```

Since you're only using the tokens for 40 minutes, maybe decrease the exp value to `issuedAt + (55 * 10)` for example?

mattiasdrp:
from one hour to 55 minutes to make sure that it's not above one hour (max allowed by GitHub)

Fixes #204

### DIFF
--- a/bot-components/GitHub_app.ml
+++ b/bot-components/GitHub_app.ml
@@ -23,7 +23,7 @@ let make_jwt ~key ~app_id =
   let issuedAt = Unix.time () |> Int.of_float in
   let payload =
     f "{ \"iat\": %d, \"exp\": %d, \"iss\": %d }" issuedAt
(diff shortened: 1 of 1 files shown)


Zimmi48 (approved):
Thanks! I'm pretty sure that the margin is still sufficient, so I will merge this immediately.

PR #206, opened 2022-03-07
read-strings is preferred to read 

mattiasdrp:
Since read will read the newline and some editors add a newline at the end of files


mattiasdrp:
Yeah, it annoyed me sufficiently to take the time to find how I could fix it :D

PR #207, opened 2022-03-07
Strings that need to use double quotes are now enclosed in {| |} for readability

mattiasdrp:
This is a QoL change because I find it hard to read and debug strings that have `\"` instead of simply `"`

### DIFF
--- a/bot-components/GitHub_mutations.ml
+++ b/bot-components/GitHub_mutations.ml
@@ -202,7 +202,7 @@ let update_milestone ~bot_info new_milestone (issue : issue) =
     |> Uri.of_string
   in
   let body =
-    {|{"milestone": |} ^ new_milestone ^ "}" |> Cohttp_lwt.Body.of_string
+    f {|{"milestone": %s}|} new_milestone |> Cohttp_lwt.Body.of_string
   in
   Lwt_io.printf "Sending patch request.\n"
   >>= fun () -> Client.patch ~headers ~body uri >>= print_response
@@ -228,8 +228,7 @@ let send_status_check ~bot_info ~repo_full_name ~commit ~state ~url ~context
 
 let add_pr_to_column ~bot_info ~pr_id ~column_id =
   let body =
-    {|{"content_id":|} ^ Int.to_string pr_id
-    ^ {|, "content_type": "PullRequest"}|}
+    f {|{"content_id":%d, "content_type": "PullRequest"}|} pr_id
     |> (fun body ->
          Stdio.printf "Body:\n%s\n" body ;
          body )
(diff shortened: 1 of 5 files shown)


bot-components/GitHub_mutations.ml
```diff
@@ -201,9 +201,9 @@ let send_status_check ~bot_info ~repo_full_name ~commit ~state ~url ~context
     repo_full_name commit state
   >>= fun () ->
   let body =
-    "{\"state\": \"" ^ state ^ "\",\"target_url\":\"" ^ url
-    ^ "\", \"description\": \"" ^ description ^ "\", \"context\": \"" ^ context
-    ^ "\"}"
+    {|{"state": "|} ^ state ^ {|","target_url":"|} ^ url
+    ^ {|", "description": "|} ^ description ^ {|", "context": "|} ^ context
+    ^ {|"}|}
```
Zimmi48, line 206: This begs the question: isn't there a good antiquotation mechanism in OCaml that would make this string even more readable (we could always rely on `Printf.fprintf`, which I have abbreviated into `f` in this code base, of course, but it's not terribly satisfying).

Zimmi48 (approved):
Great suggestion! Unfortunately, I have pushed some concurrent changes that have created a merge conflict. Do you want to take care of resolving it or should I?

bot-components/GitHub_mutations.ml
mattiasdrp: I wanted to talk about that too :D 
The thing is, if we use `sprintf` and write the arguments in the wrong order, the bugs will be really hard to fine so the current solution, even though hard to read, is the best currently in my opinion.

I'm waiting for named parameters in formatting strings that would allow us to write:

`{|{"state": "{state}", "target_url":"{url}", "description": "{description}", "context": "{context}"}|}`

But in the meantime, I'd stick to this

mattiasdrp:
I'll rebase my PR

mattiasdrp:
Rebase done

bot-components/GitHub_mutations.ml
```diff
@@ -202,7 +202,7 @@ let update_milestone ~bot_info new_milestone (issue : issue) =
     |> Uri.of_string
   in
   let body =
-    "{\"milestone\": " ^ new_milestone ^ "}" |> Cohttp_lwt.Body.of_string
+    {|{"milestone": |} ^ new_milestone ^ "}" |> Cohttp_lwt.Body.of_string
```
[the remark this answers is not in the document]
mattiasdrp: I didn't know if you wanted this, I'll do it ;-)

JasonGross:
This PR broke CI minimization because `{|...|}` strings are literal.  [The documentation says](https://ocaml.org/manual/lex.html#sss:stringliterals):
> Quoted strings do not interpret any character in a special way but requires that the sequence `|quoted-string-id}` does not occur in the string itself.

I will prepare a PR reverting most of this commit.  I'll leave any changes to strings that have no special characters and do not span multiple lines (as even the `\` followed by a newline does not get interpreted correctly, and I think the change in indentation required for using `{|....|}` is more ugly than escaping quotes).

PR #208, opened 2022-03-14
Revert most of #207

JasonGross:
This reverts commit 48fcdf2f4bbb5fc3ecab1c94354c23129e0edb1f.

#207 broke CI minimization (see https://github.com/coq/coq/pull/15780#issuecomment-1066888774) because
`{|...|}` strings are literal.  [The documentation says](https://ocaml.org/manual/lex.html#sss:stringliterals):
> Quoted strings do not interpret any character in a special way but requires that the sequence `|quoted-string-id}` does not occur in the string itself.

### DIFF
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -66,9 +66,9 @@ let send_status_check ~bot_info job_info ~pr_num (gh_owner, gh_repo)
           |> String.concat ~sep:"\n" )
     | Some index_of_error ->
         ( f
-            {|We show below an excerpt from the trace from GitLab starting \
-             around the last detected "Error" (the complete trace is \
-             available [here](%s)).|}
+            "We show below an excerpt from the trace from GitLab starting \
+             around the last detected \"Error\" (the complete trace is \
+             available [here](%s))."
             job_url
         , trace_lines
           |> Fn.flip List.drop (index_of_error - 1)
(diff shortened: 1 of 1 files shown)


Zimmi48:
Indeed, here is an example broken check run output:
https://github.com/coq/coq/pull/15807/checks?check_run_id=5538482803
And of a broken comment:
https://github.com/coq/coq/pull/15434#issuecomment-1062489482

So the lesson is that `{| |}` should not be used with multiline strings?

Zimmi48:
> So the lesson is that `{| |}` should not be used with multiline strings?

So indeed, for multiline strings, using this properly would make the code a bit weird, to say the least:

```ocaml
        f
          {|This job ran on the Docker image `%s`, depended on the build job `%s` with OCaml `%s`.

|}
          docker_image build_dependency switch_name
```

On the other hand, for any string that was broken up on several lines by ocamlformat but does not contain any actual newline, using `{| |}` could be a way to prevent ocamlformat to break the string on multiple lines.

cc @mattiasdrp

Zimmi48 (approved):
Approving just for the sake of letting Jason merge as is if he wants to.

Zimmi48:
Quoting from the comment on the other PR (I've only seen this now).

> I will prepare a PR reverting most of this commit. I'll leave any changes to strings that have no special characters and do not span multiple lines (as even the `\` followed by a newline does not get interpreted correctly, and I think the change in indentation required for using `{|....|}` is more ugly than escaping quotes).

I agree about the ugliness in the change in indentation. I guess the only thing that we really need to discuss is the case of strings that do not contain any actual newline.

JasonGross:
> So the lesson is that `{| |}` should not be used with multiline strings?

Not just multiline strings, but also other escape characters (e.g., tabs).  And also that if you convert a string like `"\\(\"\\)"` to use `{| ... |}`, you need to do `{|\("\)|}` instead of `{|\\("\\)|}`

mattiasdrp:
Oh indeed:

```ocaml
utop # let qs = {|test\
blah|};;
val qs : string = "test\\\nblah"
```

And OCamlformat doesn't create multiple lines when handling quoted strings. I never noticed. Today I learned, that's a good day :-)
Thanks for this PR, @JasonGross , and sorry for the disturbance

Zimmi48:
> Not just multiline strings, but also other escape characters (e.g., tabs). And also that if you convert a string like `"\\(\"\\)"` to use `{| ... |}`, you need to do `{|\("\)|}` instead of `{|\\("\\)|}`

Good point about the regexp strings, but this would actually make them more readable probably.

PR #211, opened 2022-03-16
Add bench command for coqbot

Alizter:
We can now tell coqbot to start the bench in a PR with `@coqbot bench`.

I haven't tested the regex fully so we need to try it out before deploying fully like usual.

### DIFF
--- a/bot-components/GitHub_queries.ml
+++ b/bot-components/GitHub_queries.ml
@@ -777,8 +777,6 @@ let get_base_and_head_checks ~bot_info ~owner ~repo ~pr_number ~base ~head =
               | _ ->
                   Error "Got more than one checkSuite." ) )
 
-(* TODO: use GraphQL API *)
-
 let get_pipeline_summary ~bot_info ~owner ~repo ~head =
   let appId = bot_info.app_id in
   let open GitHub_GraphQL.GetPipelineSummary in
@@ -835,6 +833,8 @@ let get_pipeline_summary ~bot_info ~owner ~repo ~head =
                       Error "Got more than one checkRun." ) )
               | _ ->
                   Error "Got more than one checkSuite." ) )
+
+(* TODO: use GraphQL API *)
 
 let get_cards_in_column column_id =
   generic_get
--- a/bot-components/GitHub_queries.ml
+++ b/bot-components/GitHub_queries.ml
@@ -779,6 +779,63 @@ let get_base_and_head_checks ~bot_info ~owner ~repo ~pr_number ~base ~head =
 
 (* TODO: use GraphQL API *)
 
+let get_pipeline_summary ~bot_info ~owner ~repo ~head =
+  let appId = bot_info.app_id in
+  let open GitHub_GraphQL.GetPipelineSummary in
+  makeVariables ~appId ~owner ~repo ~head ()
+  |> serializeVariables |> variablesToJson
+  |> GraphQL_query.send_graphql_query ~bot_info ~query
+       ~parse:(Fn.compose parse unsafe_fromJson)
+  >|= Result.bind ~f:(fun resp ->
+          resp.repository
+          |> Result.of_option ~error:(f "Unknown repository %s/%s." owner repo) )
+  >|= Result.bind ~f:(fun repository ->
+          match repository.getPipelineSummaryCommit with
+          | None ->
+              Error (f "Unknown head commit %s/%s@%s." owner repo head)
+          | Some (`UnspecifiedFragment _) ->
+              Error "Unspecified fragment."
+          | Some (`Commit head_obj) -> (
+              let extract_check_suites checkSuites =
+                checkSuites
+                |> Option.bind ~f:(fun checkSuites -> checkSuites.nodes)
+                |> Option.map ~f:Array.to_list
+                |> Option.map ~f:List.filter_opt
+              in
+              match extract_check_suites head_obj.checkSuites with
+              | None ->
+                  Error
+                    (f "No check suite %d for head commit %s/%s@%s." appId owner
+                       repo head )
+              | Some [headCheckSuite] -> (
+                match headCheckSuite.getPipelineSummaryCheckRuns with
+                | None ->
+                    Error
+                      (f
+                         "No GitLab CI pipeline summary for head commit \
+                          %s/%s@%s."
+                         owner repo head )
+                | Some checkRuns -> (
+                  match checkRuns.nodes |> Option.map ~f:Array.to_list with
+                  | None | Some [None] ->
+                      Error
+                        (f
+                           "No GitLab CI pipeline summary for head commit \
+                            %s/%s@%s."
+                           owner repo head )
+                  | Some [Some headCheckRun] ->
+                      Stdlib.Option.to_result
+                        ~none:
+                          (f
+                             "No GitLab CI pipeline summary for head commit \
+                              %s/%s@%s."
+                             owner repo head )
+                        headCheckRun.summary
+                  | Some _ ->
+                      Error "Got more than one checkRun." ) )
+              | _ ->
+                  Error "Got more than one checkSuite." ) )
+
 let get_cards_in_column column_id =
   generic_get
     ("projects/columns/" ^ Int.to_string column_id ^ "/cards")
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -2499,9 +2499,10 @@ let coq_check_stale_pr ~bot_info ~owner ~repo ~after ~throttle =
 let run_bench ~bot_info comment_info =
   (* Do we want to use this more often? *)
   let open Lwt.Syntax in
-  let owner = comment_info.issue.issue.owner in
-  let repo = comment_info.issue.issue.repo in
-  let pr_number = comment_info.issue.number in
+  let pr = comment_info.issue in
+  let owner = pr.issue.owner in
+  let repo = pr.issue.repo in
+  let pr_number = pr.number in
   (* We need the GitLab build_id and project_id. Currently there is no good way
      to query this data so we have to jump through some somewhat useless hoops in
      order to get our hands on this information. TODO: do this more directly.*)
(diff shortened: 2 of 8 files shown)


src/actions.ml
```diff
@@ -2288,3 +2288,88 @@ let coq_check_stale_pr ~bot_info ~owner ~repo ~after ~throttle =
...
+  in
+  match gitlab_check_summary with
+  | Error err ->
+      Lwt_io.print (f "Error: %s\n" err)
+  | Ok summary -> (
+    try
+      let build_id =
+        let regex =
+          Str.regexp
+            "/\\[bench\\]\\(https:\\/\\/gitlab\\.com\\/.*\\/jobs\\/\\([0-9]*\\)"
```
JasonGross, line 2359: Maybe you want to escape less by using `{|...|}`?
```suggestion
            {|/\[bench\]\(https:\/\/gitlab\.com\/.*\/jobs\/\([0-9]*\)|}
```
But also your regex is not well balanced.  Maybe you want to do something like `f {|%s.*\/jobs\/\([0-9]*\)|} (Str.quote "[bench](https://gitlab.com/")` (or whatever the regex string quoting mechanism is)?  Also, what's with the unescaped leading slash?

JasonGross: Also, why do you have a group of you're not using it?  Or does the url literally contain parentheses around the numbers?

src/actions.ml
```diff
@@ -2288,3 +2288,88 @@ let coq_check_stale_pr ~bot_info ~owner ~repo ~after ~throttle =
...
+        in
+        ( if Str.string_match regex summary 0 then Str.matched_string summary
+        else raise @@ Stdlib.Failure "Could not find GitLab bench job ID" )
+        |> Stdlib.int_of_string
+      in
+      let project_id =
+        let regex = Str.regexp "GitLab Project ID: [0-9]*" in
+        ( if Str.string_match regex summary 0 then Str.matched_string summary
```
JasonGross, line 2367: This seems wrong, you probably want a group around the numbers and then to get group 1(?), not to get the entire matched string (which will not be an integer)

src/actions.ml
```diff
@@ -2288,3 +2288,88 @@ let coq_check_stale_pr ~bot_info ~owner ~repo ~after ~throttle =
...
+            | _ ->
+                Lwt.return_error
+                  (f
+                     "Error while detecting GitLab check for %s/%s#%d for \
+                      running bench job."
+                     owner repo pr_number ) ) )
+  in
+  match gitlab_check_summary with
+  | Error err ->
+      Lwt_io.print (f "Error: %s\n" err)
```
JasonGross, line 2353: I think you want
```suggestion
      Lwt_io.printlf "Error: %s" err
```

src/actions.ml
```diff
@@ -2288,3 +2288,88 @@ let coq_check_stale_pr ~bot_info ~owner ~repo ~after ~throttle =
...
+let run_bench ~bot_info comment_info =
+  (* Do we want to use this more often? *)
+  let open Lwt.Syntax in
+  let owner = comment_info.issue.issue.owner in
+  let repo = comment_info.issue.issue.repo in
+  let pr_number = comment_info.issue.number in
+  (* We need the GitLab build_id and project_id. Currently there is no good way
+     to query this data so we have to jump through some somewhat useless hoops in
+     order to get our hands on this information. TODO: do this more directly.*)
+  let* gitlab_check_summary =
```
JasonGross, line 2301: Why `let *`?

src/actions.ml
```diff
@@ -2288,3 +2288,88 @@ let coq_check_stale_pr ~bot_info ~owner ~repo ~after ~throttle =
...
+        | Error err ->
+            Lwt.return_error
+              (f
+                 "Error while fetching checks for %s/%s#%d for running bench \
+                  job: %s"
+                 owner repo pr_number err )
+        | Ok info -> (
+            let head_checks = info.head_checks in
+            let gitlab_check =
+              List.find head_checks ~f:(function
```
JasonGross, line 2326: Do you want to use feed and function rather than let-in, to make the locality of the binding more clear?

JasonGross (changes requested):
The regexes are broken but the rest of the code looks good

src/actions.ml
Alizter: That comes from `Lwt.Syntax`, it just does `Lwt.bind` i.e. `>>=`. This let's you write monadic code slightly more cleanly, but here it hasn't really gained us much. The trouble is we only have `let*` for `Lwt.t`, but we would also need something similar for `Result.result`, but I didn't really want to have to use `ppx_let`.

src/actions.ml
Alizter: The url looks like this:
```
https://gitlab.com/coq/coq/-/jobs/2202936719
```
I'm just failing to write regex

Alizter: Ok I have fixed it now.

src/actions.ml
Alizter: Fixed

src/bot.ml
```diff
@@ -307,6 +307,22 @@ let callback _conn req body =
                 Server.respond_string ~status:`OK
                   ~body:(f "Received a request to merge the PR.")
                   () )
+              else if
+                string_match ~regexp:(f "@%s:? [Bb]ench" bot_name) body
```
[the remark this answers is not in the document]
Alizter: Done.

src/actions.ml
Alizter: Done.

src/actions.ml
```diff
@@ -2288,3 +2288,86 @@ let coq_check_stale_pr ~bot_info ~owner ~repo ~after ~throttle =
...
+      let project_id =
+        let regex = Str.regexp {|GitLab Project ID: \([0-9]*\)|} in
+        ( if Str.string_match regex summary 0 then Str.matched_group 1 summary
+        else raise @@ Stdlib.Failure "Could not find GitLab Project ID" )
+        |> Int.of_string
+      in
+      GitLab_mutations.play_job ~bot_info ~project_id ~build_id
+    with Stdlib.Failure s ->
+      Lwt_io.print
+        (f "Error while regexing summary for %s/%s#%d for running bench job: %s"
```
JasonGross, line 2372: Do you want to use printlf here instead?

Alizter: Yep.

src/actions.ml
```diff
@@ -2288,3 +2288,86 @@ let coq_check_stale_pr ~bot_info ~owner ~repo ~after ~throttle =
...
+        Lwt.return_error
+          (f
+             "Error while fetching PR refs for %s/%s#%d for running bench job: \
+              %s"
+             owner repo pr_number err )
+    | Ok {base= {sha= base}; head= {sha= head}} -> (
+        let base = Str.global_replace (Str.regexp {|"|}) "" base in
+        let head = Str.global_replace (Str.regexp {|"|}) "" head in
+        GitHub_queries.get_base_and_head_checks ~bot_info ~owner ~repo
+          ~pr_number ~base ~head
```
Zimmi48, line 2315: For testing, this is OK, but for merging the feature, it would be preferable to add a custom query that allows you to get directly the checks for the head of the PR (and not the base).

Zimmi48 (commented):
LGTM
@JasonGross If you are fine with the current state and that you have time, can you test deploy?

JasonGross:
I'm fine with the current state but don't have time to test deploy currently, sorry

Zimmi48:
@Alizter Deploy in progress at: https://github.com/Zimmi48/bot/actions/runs/2003289788

src/actions.ml
```diff
@@ -2288,3 +2288,86 @@ let coq_check_stale_pr ~bot_info ~owner ~repo ~after ~throttle =
...
+      Lwt_io.printlf "Error: %s\n" err
+  | Ok summary -> (
+    try
+      let build_id =
+        let regex =
+          Str.regexp
+          @@ f {|%s.*\/jobs\/\([0-9]*\)|}
+               (Str.quote "[bench](https://gitlab.com/")
+        in
+        ( if Str.string_match regex summary 0 then Str.matched_group 1 summary
```
Zimmi48, line 2359: Here you are using `Str.string_match` directly, but in the minimizer code, we use the `string_match` function from `Helpers` which is actually defined in terms of `Str.search_forward` and not of `Str.string_match`. This might explain the issue with not finding the result when there are newlines.

Alizter: Thank you so much!

src/actions.ml
Alizter: However `GitHub_queries.get_base_and_head_checks` is quite a large and complicated query. I am not really sure how to break it down into a smaller one.

Zimmi48: I can take care of that.

Alizter:
I've successfully deployed, feel free to merge.

JasonGross:
(not necessarily for this PR) Should we have coqbot thumbs up the comment requesting the bench once it starts the bench (a la dependabot)?  Otherwise it seems there's not feedback that coqbot has interpreted the message correctly.

Alizter:
@JasonGross I agree that the feedback is lacking. My hope was that the checks tab would be enough. I considered adding a comment but that would be noisy IMO. I quite like the thumbs up idea however, I will look into it.

Zimmi48:
I like the thumbs up idea especially because it can be generalized to a framework for any comment-based comment. (And we can also react with confused emoji when we detected a command that we could not understand.)

Zimmi48:
> I like the thumbs up idea especially because it can be generalized to a framework for any comment-based command. (And we can also react with confused emoji when we detected a command that we could not understand.)

This "framework" for comment-based commands could also be the one taking care of checking that the commenter issuing the comment is authorized (belongs to the appropriate team if this command is limited to a given team). By keeping this (security-sensitive) code in a single place, this would help avoid doing mistakes when copy-pasting / adapting it for new use cases. For instance, in the case of the "coqbot merge now" command, we have also decided to forbid commands sent by email comments. It could also make sense to follow this policy here.

bot-components/GitHub_queries.ml
```diff
@@ -779,6 +779,63 @@ let get_base_and_head_checks ~bot_info ~owner ~repo ~pr_number ~base ~head =
 
 (* TODO: use GraphQL API *)
```
Zimmi48, line 780: Oups, I should have inserted this new function above this comment, not below.

src/actions.ml
```diff
@@ -2495,3 +2495,81 @@ let coq_check_stale_pr ~bot_info ~owner ~repo ~after ~throttle =
...
+      GitLab_mutations.play_job ~bot_info ~project_id ~build_id
+  | Error err, _ | _, Error err ->
+      GitHub_mutations.post_comment ~bot_info ~message:err ~id:pr.id
+      >>= GitHub_mutations.report_on_posting_comment
+  | Ok false, _ ->
+      (* User not found in the team *)
+      let err =
+        f
+          "@%s: You can't request a bench because you're not a member of the \
+           `@coq/pushers` team."
```
Zimmi48, line 2571: FTR, my suggestion was to allow any member of `@coq/contributors` to start the bench. But this is not a blocking comment.

Zimmi48 (approved):

src/actions.ml
Alizter: We could do that actually.

Alizter:
```suggestion
  let* allowed_to_bench =
    GitHub_queries.get_team_membership ~bot_info ~org:"coq" ~team:"contributors"
      ~user:comment_info.author
  in
  match (allowed_to_bench, process_summary) with
  | Ok true, Ok (build_id, project_id) ->
      (* Permission to bench has been granted *)
      GitLab_mutations.play_job ~bot_info ~project_id ~build_id
  | Error err, _ | _, Error err ->
      GitHub_mutations.post_comment ~bot_info ~message:err ~id:pr.id
      >>= GitHub_mutations.report_on_posting_comment
  | Ok false, _ ->
      (* User not found in the team *)
      let err =
        f
          "@%s: You can't request a bench because you're not a member of the \
           `@coq/contributors` team."
```

PR #212, opened 2022-03-18
thunk merge_pull_request since optional args cannot be deleted otherwise

Alizter:
I'm trying this on a newer version of OCaml (4.12.1) and there is a warning about optional arguments that cannot be erased. Having a look at GitHub_mutations.merge_pull_requests there is no way for OCaml to understand a partial application. In order to fix this we just need to add a thunk.


Zimmi48 (approved):
Indeed, thanks for noticing and fixing this.

PR #214, opened 2022-03-21
Add bench to check status

Alizter:
When the bench job starts and and finishes we push the check status for it. I am not sure how to do it for a failure, so perhaps I should forgo the status checking all together?

### DIFF
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -257,6 +257,40 @@ let send_doc_url ~bot_info ~github_repo_full_name job_info =
   | _ ->
       Lwt.return_unit
 
+let update_bench_status ~bot_info job_info (gh_owner, gh_repo) ~external_id =
+  GitHub_queries.get_repository_id ~bot_info ~owner:gh_owner ~repo:gh_repo
+  >>= function
+  | Ok repo_id -> (
+      Lwt_io.print "Pushing status check for bench job.\n"
+      <&>
+      let url = f "https://gitlab.com/coq/coq/-/jobs/%d" job_info.build_id in
+      let state = job_info.build_status in
+      let context = "bench" in
+      ( match state with
+      | "success" ->
+          Lwt.return_ok (COMPLETED, Some SUCCESS, "Bench completed successfully")
+      | "running" ->
+          Lwt.return_ok (IN_PROGRESS, None, "Bench in progress")
+      | "cancelled" | "canceled" ->
+          Lwt.return_ok (COMPLETED, Some CANCELLED, "Bench has been cancelled")
+      | "failed" ->
+          Lwt.return_ok
+            (COMPLETED, Some NEUTRAL, "Bench completed with failures")
+      | "created" ->
+          Lwt.return_error ("Bench job has been created, ignoring info update.")
+      | _ ->
+          Lwt.return_error (f "Unknown state for bench job: %s" state) )
+      >>= function
+      | Ok (status, conclusion, description) ->
+          let description = description in
+          GitHub_mutations.create_check_run ~bot_info ~name:context ~status
+            ~repo_id ~head_sha:job_info.common_info.head_commit ?conclusion
+            ~title:description ~details_url:url ~summary:"" ~external_id ()
+      | Error e ->
+          Lwt_io.print e )
+  | Error e ->
+      Lwt_io.printlf "No repo id for bench job: %s" e
+
 type build_failure = Warn of string | Retry | Ignore
 
 let trace_action ~repo_full_name trace =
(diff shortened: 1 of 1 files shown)


SkySkimmer:
Failing benches already add a allowed_failure status

Alizter:
That's correct, so this probably will introduce some duplication as it stands.

Alizter:
The bench status should now display cancellation, failure etc correctly. There is still some duplication between the failure jobs, since if the bench is reported as a failure then it will appear as a CI job. I think we can fix this by just checking for the bench in the relevant areas. Currently this + #211 is deployed.

JasonGross:
Can the successful bench post the bench summary automatically?

Alizter:
@JasonGross That's a third thing that I want to work on, which is the posting of the bench results once the bench has been completed. For now we only have the indicator in the checks tabs.

Zimmi48:
Ah, for the record, my suggestion was not to duplicate the status reporting logic for the bench job. It was to detect if this was the bench job and activate the reporting in this case. Basically, this would have required generalizing the `job_success_or_pending` function a bit.

Alizter:
@Zimmi48 The logic looks very similar but it isn't an exact duplicate. I initially tried to generalize job_success_or_pending but it became too awkward. The thinking I have is that the bench job is special amongst the others since it never "fails" and is immediately shown. Currently we can even show it being cancelled which job_action does not support generally.

Zimmi48:
OK, then it should just be filtered out from the `job_action` main logic (as early as possible) to avoid the duplication.

Zimmi48:
There's also a bug with this PR which is that it reports about the bench as soon as the job is "created" even if it is not triggered. See example at: https://github.com/coq/coq/pull/15844. We should figure out which event to account for and which events to ignore.

src/actions.ml
```diff
@@ -257,6 +257,79 @@ let send_doc_url ~bot_info ~github_repo_full_name job_info =
...
+          | "success" ->
+              ( "success"
+              , COMPLETED
+              , Some SUCCESS
+              , "Bench completed successfully" )
+          | "created" ->
+              ( "pending"
+              , QUEUED
+              , None
+              , "Bench requested" )
```
Zimmi48, line 285: I'm guessing just removing this should be enough to fix this issue:
```suggestion
```

Zimmi48: But possibly a better option would just be to replace "created" with "pending":
```suggestion
          | "pending" ->
              ( "pending"
              , QUEUED
              , None
              , "Bench requested" )
```

Currently, the bot log contains a lot of "Unknown job status: pending"

Alizter: Another bug is that a bench hasn't even been requested. I think I will settle for not including this.

src/actions.ml
```diff
@@ -257,6 +257,55 @@ let send_doc_url ~bot_info ~github_repo_full_name job_info =
...
+      | "created" ->
+          ("pending", QUEUED, None, "Bench ready — Comment \"@coqbot bench\" to start bench")
+      | "running" ->
+          ("pending", IN_PROGRESS, None, "Bench in progress")
+      | "cancelled" | "canceled" ->
+          ("failure", COMPLETED, Some CANCELLED, "Bench cancelled")
+      | "failed" ->
+          ("failure", COMPLETED, Some NEUTRAL, "Bench completed with failures")
+      | _ ->
+          failwith (f "Error: send_bench_url received unknown state %s." state)
```
Zimmi48, line 284: You should not raise an anomaly for something as non-problematic like this, just print a warning as in the `job_action` function. And you should not even use the "Error" keyword. The fact that this keyword is pretty rare is useful so that we can retrieve the issues quickly.

Alizter: Right this is copied from job_action.

Alizter: I will remove it.

Alizter: Sorry not job_action but success or pending

src/actions.ml
Zimmi48: >Another bug is that a bench hasn't even been requested

This was my point actually :wink:

src/actions.ml
Zimmi48: Yes, but `success_or_pending` is only called on pre-filtered states, so the failwith is only there to avoid getting an "incomplete pattern matching" warning (it is a true "impossible state").

Alizter: What can I put here instead? Should I raise and catch an exception instead?

Zimmi48: You should do a simple print, like the:

```ocaml
    | unknown_state ->
        Lwt_io.printf "Unknown job status: %s\n" unknown_state )
```

in the `job_action` function.

Alizter: The issue is that this branch is returning 4 values and not an Lwt.

Zimmi48: Right. I missed that. Then, my recommendation is to wrap the result in a result monad. That being said, you could also refactor the code of this function to return an Lwt instead.

Alizter: OK i've raised and handled an exception instead.

src/actions.ml
```diff
@@ -257,6 +257,55 @@ let send_doc_url ~bot_info ~github_repo_full_name job_info =
...
+          ~description ~bot_info
+    | Some _ -> (
+        GitHub_queries.get_repository_id ~bot_info ~owner:gh_owner ~repo:gh_repo
+        >>= function
+        | Ok repo_id ->
+            GitHub_mutations.create_check_run ~bot_info ~name:context ~status
+              ~repo_id ~head_sha:job_info.common_info.head_commit ?conclusion
+              ~title:description ~details_url:url ~summary:"" ~external_id ()
+        | Error e ->
+            Lwt_io.printf "No repo id: %s\n" e )
```
Zimmi48, line 302: You can consider that for the Coq repository, we are always in the `Some` case. So the `get_repository_id` request could be factorized at the beginning, then you would call `create_check_run` directly for each type of change.

src/actions.ml
```diff
@@ -257,6 +257,55 @@ let send_doc_url ~bot_info ~github_repo_full_name job_info =
...
+let send_bench_url ~bot_info job_info (gh_owner, gh_repo) ~external_id =
+  Lwt_io.print "Pushing status check for bench job.\n"
+  <&>
+  let url = f "https://gitlab.com/coq/coq/-/jobs/%d" job_info.build_id in
+  let state = job_info.build_status in
+  let context = "bench" in
+  url |> Uri.of_string |> Client.get
+  >>= fun (resp, _) ->
+  let status_code = resp |> Response.status |> Code.code_of_status in
+  if Int.equal 200 status_code then
```
Zimmi48, line 269: FTR, I get it that you've copied this code from `send_doc_url`, but it is completely irrelevant for this function: in the case of `send_doc_url`, we check that the artifact was correctly deployed. Here, we do not need to do this because there is no artifact that we point to. Even the name of the function `send_bench_url` is a bit misleading in this regard.

Alizter: Ok i've changed the name to update_bench_status and removed all this stuff.

src/actions.ml
Alizter: Removed all of it now.

Zimmi48 (approved):
LGTM

src/actions.ml
```diff
@@ -257,6 +257,43 @@ let send_doc_url ~bot_info ~github_repo_full_name job_info =
...
+      let url = f "https://gitlab.com/coq/coq/-/jobs/%d" job_info.build_id in
+      let state = job_info.build_status in
+      let context = "bench" in
+      ( match state with
+      | "success" ->
+          Lwt.return_ok (COMPLETED, Some SUCCESS, "Bench completed successfully")
+      | "created" ->
+          Lwt.return_ok
+            ( QUEUED
+            , None
```
Alizter, line 275: This is why nothing is shown when the checks are run. Should we change this to `Some NEUTRAL`?

Alizter: By the way, `NEUTRAL` will get completely ignored if we also don't change `QUEUED` to `COMPELTED` IIUC. This is what I have done. That's a good point about neutral however.

Alizter: I tried action_requires but that fails the check which isn't great. I've switched back to neutral.

Alizter: Which by the way is all happening in the other PR not this one. I think we can merge this one now @Zimmi48.

Alizter: Unless we want to make the neutral happen here...

PR #217, opened 2022-03-22
Bench report

Alizter:
Depends on #214

Here it is in actions: https://github.com/coq/coq/pull/14748/checks?check_run_id=5651821583

### DIFF
--- a/bot-components/GitHub_mutations.ml
+++ b/bot-components/GitHub_mutations.ml
@@ -58,8 +58,8 @@ let close_pull_request ~bot_info ~pr_id =
   | Error err ->
       Stdio.print_endline (f "Error while closing PR: %s" err)
 
-let merge_pull_request ~bot_info ?merge_method ?commit_headline
-    ?commit_body ~pr_id () =
+let merge_pull_request ~bot_info ?merge_method ?commit_headline ?commit_body
+    ~pr_id () =
   let merge_method =
     Option.map merge_method ~f:(function
       | MERGE ->
@@ -150,11 +150,11 @@ let create_check_run ~bot_info ?conclusion ~name ~repo_id ~head_sha ~status
   |> serializeVariables |> variablesToJson
   |> GraphQL_query.send_graphql_query ~bot_info ~query
        ~parse:(Fn.compose parse unsafe_fromJson)
-  >|= function
-  | Ok _ ->
-      ()
-  | Error err ->
-      Stdio.print_endline (f "Error while creating check run: %s" err)
+  >|= Result.bind ~f:(function
+        | {createCheckRun= Some {checkRun= Some {url}}} ->
+            Ok url
+        | _ ->
+            Error (f "Error while creating check run.") )
 
 let update_check_run ~bot_info ~check_run_id ~repo_id ~conclusion ?details_url
     ~title ?text ~summary () =
--- a/bot-components/GitHub_mutations.ml
+++ b/bot-components/GitHub_mutations.ml
@@ -154,7 +154,7 @@ let create_check_run ~bot_info ?conclusion ~name ~repo_id ~head_sha ~status
         | {createCheckRun= Some {checkRun= Some {url}}} ->
             Ok url
         | _ ->
-            Error (f "Warning: Check run could not be created.") )
+            Error (f "Warning: Could not retrieve new check run URL.") )
 
 let update_check_run ~bot_info ~check_run_id ~repo_id ~conclusion ?details_url
     ~title ?text ~summary () =
--- a/bot-components/GitHub_mutations.ml
+++ b/bot-components/GitHub_mutations.ml
@@ -154,7 +154,7 @@ let create_check_run ~bot_info ?conclusion ~name ~repo_id ~head_sha ~status
         | {createCheckRun= Some {checkRun= Some {url}}} ->
             Ok url
         | _ ->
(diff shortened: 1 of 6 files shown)


Alizter:
depends on #222

src/actions.ml
```diff
@@ -331,47 +329,111 @@ let bench_text ~job_info () =
...
-          (COMPLETED, Some CANCELLED, "Bench has been cancelled", "")
-      | "created" ->
-          Lwt.return_error ("Bench job has been created, ignoring info update.")
-      | _ ->
-          Lwt.return_error (f "Unknown state for bench job: %s" state) )
+  match number with
+  | None ->
+      Lwt_io.printlf "No PR number provided for bench summary so aborting."
+  | Some number -> (
+      GitHub_queries.get_repository_id ~bot_info ~owner:gh_owner ~repo:gh_repo
```
Zimmi48, line 381: Actually, in the cases where we need both the repository id and the PR id, instead of doing two requests, we could only do the latter and amend it to also return the repository id.

src/actions.ml
```diff
@@ -331,47 +329,111 @@ let bench_text ~job_info () =
...
+              let* summary_text = bench_text results in
+              let* () =
+                bench_comment ~bot_info ~owner:gh_owner ~repo:gh_repo ~number
+                  results
+              in
+              Lwt.return_ok
+                ( COMPLETED
+                , Some NEUTRAL
+                , "Bench completed with failures"
+                , summary_text )
```
Zimmi48, line 416: I'm really not fond of this logic where part of the action is done inside this match and part of it is done outside, using the return value. My initial suggestion was to do the action inside the match, and it still holds.

In fact, I have a further suggestion which is to put a link from the comment to the check run summary. Following it would require reversing the order in which the actions are done: the check run would need to be created first, and amended to return its URL. Then the comment could be created and use this piece of data.

Alizter: Let me try to factor things out from the match then. I agree that the flow isn't great at the moment.

Zimmi48: Do as you want (and maybe it will look great) but note that my suggestion was to put everything inside the match instead.

Alizter: I've put everything in the match now.

src/actions.ml
```diff
@@ -384,64 +385,66 @@ let update_bench_status ~bot_info job_info (gh_owner, gh_repo) ~external_id
...
             | Ok url ->
-                Lwt_io.printlf "Bench Check Summary updated: %s" url
+                let* () =
+                  Lwt_io.printlf "Bench Check Summary updated: %s" url
+                in
+                Lwt.return_some url
             | Error e ->
-                Lwt_io.printf "Bench Check Summary url missing: %s" e
+                let* () =
+                  Lwt_io.printf "Bench Check Summary URL missing: %s" e
```
Zimmi48, line 408:
```suggestion
                  Lwt_io.printlf "Bench Check Summary URL missing: %s" e
```

bot-components/GitHub_mutations.ml
```diff
@@ -150,11 +150,11 @@ let create_check_run ~bot_info ?conclusion ~name ~repo_id ~head_sha ~status
...
-  >|= function
-  | Ok _ ->
-      ()
-  | Error err ->
-      Stdio.print_endline (f "Error while creating check run: %s" err)
+  >|= Result.bind ~f:(function
+        | {createCheckRun= Some {checkRun= Some {url}}} ->
+            Ok url
+        | _ ->
+            Error (f "Error while creating check run.") )
```
Zimmi48, line 157: You should keep propagating the error message instead of dropping it. This would give you detailed information on how you failed to get the check run.

Alizter: I think I copied the comment case a bit too literally here. I will try to make a more specific match and propagate the error correctly like you said.

Alizter: Wait isn't this exactly what Result.bind is doing. Here we are separating the case where there is None for the url field and returning an error. The original error would be preserved.

Alizter: Very well, I will remove the link.

Zimmi48: Or maybe just keep the code so that if GitHub starts being more reactive and returning the check run URL we can see it appear in our PRs...

Alizter: OK I won't touch it then. I set it up so that it doesn't impede the main function anyway.

Zimmi48: Maybe just replace the "Error" by a "Warning" so that we do not get error messages in the log for something that we already know is not working.

bot-components/GitHub_mutations.ml
```diff
@@ -154,7 +154,7 @@ let create_check_run ~bot_info ?conclusion ~name ~repo_id ~head_sha ~status
         | {createCheckRun= Some {checkRun= Some {url}}} ->
             Ok url
         | _ ->
-            Error (f "Error while creating check run.") )
+            Error (f "Warning: Check run could not be created.") )
```
Zimmi48, line 157: Actually, as we have determined together, since we are inside a bind, it means that the mutation has still returned successfully, so what we should warn about is not that the check run could not be created but that we didn't get the URL back.

Zimmi48 (approved):

PR #226, opened 2022-05-03
Account for new pipeline status: created.

Zimmi48:
Should avoid issues like the one observed at: https://github.com/coq/coq/pull/15986/checks?check_run_id=6270508750


JasonGross (approved):
@Zimmi48 This has a conflict, do you want to rebase and merge?

Zimmi48:
I haven't seen recent cases where the pipeline was taking time to be reported as pending, so maybe that's not so useful anymore?

PR #232, opened 2022-06-23
[coq] [doc] Update for new upstream documentation jobs

ejgallego:
This should be merged after #15560

### DIFF
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -235,35 +235,27 @@ let send_doc_url_aux ~bot_info job_info (kind, url) =
       ~description:(description_base ^ ": not found.")
       ~bot_info
 
+let send_doc_url_job ~bot_info job_info doc_key artifact =
+  Lwt_io.printf
+    "This is a successful %s build. Pushing a status check with a \
+     link...\n" doc_key
+  <&>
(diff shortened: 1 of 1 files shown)


Alizter (approved):

Alizter:
Guillaume has pointed out that this will break the CI for other branches. Would it be possible to fall back to the old link if the new doesn't work?

Zimmi48:
This is exactly what I would recommend. Then, this PR could be merged first (like a backward compatible overlay).

ejgallego:
Config like this tho shouldn't live in the source code, maybe the bot could read a `bot-config.json` file from each repos?

[excuse my unfamiliarity with the bot]

ejgallego:
Not sure Ali can push here, as the box is already checked, so closing in favor of #233

Zimmi48:
> Config like this tho shouldn't live in the source code, maybe the bot could read a `bot-config.json` file from each repos?
> 
> [excuse my unfamiliarity with the bot]

We do have a feature for reading a repo-specific configuration file (`coqbot.toml`). It would just need to be extended to handle this specific feature. But that would require designing what this feature means in general (beyond the specific case of Coq's documentation artifacts).

PR #233, opened 2022-06-24
[coq] [doc] Update for new upstream documentation jobs

Alizter:


Alizter:
Currently this is deployed, we I will merge when I see a CI from before and after your PR go green.

Alizter:
Merging this can come later in the week since it is already deployed.

Alizter:
This doesn't seem to be adding the link for older refman jobs correctly. Though they are not failing.

Alizter:
@Zimmi48 This is working however it doesn't show the links for PRs before the Dune PR. They however do not fail. I would suggest merging this, and eventually we can come up with something cleaner, since the old linking isn't really working here. I'm also not sure how useful that will be.

Zimmi48:
That's fine, I guess, but do you understand why the link do not show up?

Alizter:
@Zimmi48 not really.

Zimmi48:
OK, then I want to review properly this PR and try to understand what's going on before merging.

Zimmi48:
I've rebased the PR on top of #232 for my own convenience.
The commit that needs reviewing is the one I've pushed on top of yours.
The first change is that only one fallback artifact location is provided per function call. This allows to simplify the logic quite a bit. Note also the use of recursion to handle the fallback without repeating any code.

Alizter:
Your commit LGTM. Shall we deploy and test?

Zimmi48:
Deployed and testing in progress: I restarted the following two jobs:
- https://gitlab.com/coq/coq/-/jobs/3033009951 (current head of master)
- https://gitlab.com/coq/coq/-/jobs/3032989370 (current head of v8.16)

Zimmi48:
It looks like the fallback URL works now: https://github.com/coq/coq/commit/628f33ed583d0f84a05db377b533fafe460abace
We still have to wait for the test on master, but I don't see why it would work for the fallback URL and not work for the normal case.

PR #240, opened 2022-10-16
adapt bench handling code to bench tweaks upstream

Alizter:
We need to deploy and test for https://github.com/coq/coq/pull/16680

They should be merged at the same time.

### DIFF
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -280,18 +280,7 @@ let send_doc_url ~bot_info ~github_repo_full_name job_info =
   | _ ->
       Lwt.return_unit
 
-module BenchResults = struct
-  type t =
-    { summary_table: string
-    ; slow_table: string
-    ; slow_number: int
-    ; fast_table: string
-    ; fast_number: int }
-end
-
 let fetch_bench_results ~job_info () =
-  let open BenchResults in
-  let open Lwt.Syntax in
   let fetch_artifact url =
     url |> Uri.of_string |> Client.get
     >>= fun (resp, body) ->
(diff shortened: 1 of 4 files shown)


SkySkimmer:
What's this ps-id stuff?

Alizter:
@SkySkimmer Its from the tool I use to manage my stack of patches: https://github.com/uptech/git-ps-rs

It saves me from having to create branches for everything. I just have a single master branch and tell it which patches I want to submit for review.

Zimmi48:
@Alizter I'll leave the hand to you to deploy / test this as you see fit. The currently deployed version of coqbot corresponds to the current HEAD of the `master` branch, so there should be no surprises if you want to test this. And I'll refrain from testing new things in parallel.

Alizter:
@Zimmi48 I'm not looking to test this today, I will let you know when I plan to.

src/actions.ml
```diff
@@ -304,167 +306,161 @@ let fetch_bench_results ~job_info () =
...
+        ; fast_table
+        ; slow_table_pdiff
+        ; fast_table_pdiff } =
+    results
+  in
+  [ pp_table "Bench Summary" summary_table ~emoji:":checkered_flag:"
+  ; pp_table "Slow Downs" slow_table ~emoji:":turtle:"
+  ; pp_table "Speed Ups" fast_table ~emoji:":rabbit2:"
+  ; pp_table "Slow Downs (%)" slow_table_pdiff ~emoji:":snail:"
+  ; pp_table "Speed Ups (%)" fast_table_pdiff ~emoji:":eagle:" ]
```
Zimmi48, line 356: Having looked at the [result](https://github.com/coq/coq/pull/16680#issuecomment-1282955838), I think you should make pdiff or something like that appear in the summary. The `(%)` is too easy to miss and we don't really get the difference from the previous tables.

src/actions.ml
```diff
@@ -304,167 +306,161 @@ let fetch_bench_results ~job_info () =
...
+  |> Caml.List.flatten |> String.concat ~sep:"\n"
```
Zimmi48, line 357: What is "Caml"?

src/actions.ml
```diff
@@ -304,167 +306,161 @@ let fetch_bench_results ~job_info () =
...
-        Lwt_io.printlf "Unable to fetch_results for bench for pr #%d: %s" number
-          e )
-  | Error e ->
-      Lwt_io.printlf "Unable to get_pull_request_id for bench for pr #%d: %s"
-        number e
+let bench_comment ~bot_info ~owner ~repo ~number ~gitlab_url ~check_url
+    (results : BenchResults.t) =
+  let open Lwt_result.Syntax in
+  let* id = GitHub_queries.get_pull_request_id ~bot_info ~owner ~repo ~number in
+  (* Formatting heleprs *)
```
Zimmi48, line 363:
```suggestion
  (* Formatting helpers *)
```

src/actions.ml
```diff
@@ -304,167 +306,161 @@ let fetch_bench_results ~job_info () =
...
+      @@ f "%s %a %s:" emoji pp_count table.BenchResults.Table.count title
+    ; code_wrap table.table ]
+  in
+  (* Document *)
+  let BenchResults.
+        { summary_table
+        ; slow_table
+        ; fast_table
+        ; slow_table_pdiff
+        ; fast_table_pdiff } =
```
Zimmi48, line 349: Why not just replace `result` in the argument of the function with this?

src/actions.ml
```diff
@@ -304,167 +306,161 @@ let fetch_bench_results ~job_info () =
...
+      (f "%s %a %s" emoji pp_count table.count title)
+      (code_wrap table.table)
+  in
+  (* For making sure we are using all the data *)
+  let BenchResults.
+        { summary_table
+        ; slow_table
+        ; fast_table
+        ; slow_table_pdiff
+        ; fast_table_pdiff } =
```
Zimmi48, line 389: Ditto.

Zimmi48 (approved):
Approving so that you can merge when appropriate.

SkySkimmer:
The bench reporting seems pretty fragile, could we fix it so it reports even if some of the timing files are not present?

Zimmi48:
> The bench reporting seems pretty fragile, could we fix it so it reports even if some of the timing files are not present?

I would recommend that indeed, especially given that we have already seen cases of some artifacts missing.

src/actions.ml
Alizter: It has been a while, but I believe that is Base reexporting the stdlib since it shadows List. I think flatten wasn't available.

src/actions.ml
Alizter: I wrote it like this so it is obvious that all the tables are being printed.

src/actions.ml
Zimmi48: OK. Then, an alternative way to access it that is already used in a few places is through the explicit `Stdlib` module.

src/actions.ml
Zimmi48: I've seen the comment, but what is the difference with what I propose? You would still get a warning if one of these variables is not used...

Alizter:
Currently rewriting this to be less fragile and more idiomatic.

src/actions.ml
Alizter: I've stopped using it all together now.

src/actions.ml
Alizter: I've made it more obvious now.

Zimmi48:
LGTM, but maybe worth being explicit on why there are missing tables when it's the case. Something like: "Could not retrieve all the tables. Some artifacts are missing."

PR #241, opened 2022-10-16
Add functionality to bench native

Alizter:
We should now be able to do `@coqbot bench native`. Once https://github.com/coq/coq/pull/16679 lands.

This defo needs testing since the GitLab docs on the env var for the job are so poor.

### DIFF
--- a/src/bot.ml
+++ b/src/bot.ml
@@ -332,6 +332,26 @@ let callback _conn req body =
                   () )
               else if
                 string_match
+                  ~regexp:(f "@%s:? [Bb]ench native" @@ Str.quote bot_name)
+                  body
+                && comment_info.issue.pull_request
+                && String.equal comment_info.issue.issue.owner "coq"
+                && String.equal comment_info.issue.issue.repo "coq"
+                && signed
+              then (
+                (fun () ->
+                  action_as_github_app ~bot_info ~key ~app_id
+                    ~owner:comment_info.issue.issue.owner
+                    ~repo:comment_info.issue.issue.repo
+                    (run_bench
+                       ~key_value_pairs:[("coq_native", "yes")]
+                       comment_info ) )
+                |> Lwt.async ;
+                Server.respond_string ~status:`OK
+                  ~body:(f "Received a request to start the bench.")
+                  () )
+              else if
+                string_match
                   ~regexp:(f "@%s:? [Bb]ench" @@ Str.quote bot_name)
                   body
                 && comment_info.issue.pull_request
--- a/bot-components/GitLab_mutations.ml
+++ b/bot-components/GitLab_mutations.ml
@@ -1,4 +1,6 @@
+open Base
 open Bot_info
+open Utils
 
 let generic_retry ~bot_info ~url_part =
   let uri =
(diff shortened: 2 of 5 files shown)


Zimmi48:
FWIW, I just looked into the GitLab GraphQL mutation for playing jobs in case it would provide a clearer and safer API to do this, and it does not have support for setting variables.

Zimmi48:
My reading of the [documentation](https://docs.gitlab.com/ee/api/jobs.html#run-a-job) is that this variable is expected to be passed as part of the body of the POST request. I would prefer to do this rather than URL encode the value. It seems to me that it would make the code more readable, and probably less fragile.

Alizter:
@Zimmi48 Feel free to take over and make a POST request. I do not know how to do that. I agree that the current solution is not readable at all. While you are at it, you might as well make ?key_value into ?key_values and take a list. We might add functionality to `@coqbot bench` doing `@coqbot bench only bugnums` or something. Being able to set multiple env vars would be ideal for this.

Zimmi48:
OK, I will push a commit to this PR, then hand it to you again for testing when this is the right time for you.

Zimmi48:
Done.

src/bot.ml
```diff
@@ -330,6 +330,26 @@ let callback _conn req body =
...
+                (fun () ->
+                  action_as_github_app ~bot_info ~key ~app_id
+                    ~owner:comment_info.issue.issue.owner
+                    ~repo:comment_info.issue.issue.repo
+                    (run_bench
+                       ~key_value_pairs:[("coq_native", "yes")]
+                       comment_info ) )
+                |> Lwt.async ;
+                Server.respond_string ~status:`OK
+                  ~body:(f "Received a request to start the bench.")
```
Zimmi48, line 351: Maybe this message could be tweaked to be more specific about the request received?

Zimmi48 (approved):
Feel free to merge.

src/bot.ml
Alizter: I couldn't even find it in the heroku logs

Zimmi48: This is not supposed to appear in the Heroku logs. It is the feedback that is sent to GitHub. You can find it by looking at the response of any webhook here: https://github.com/organizations/coq/settings/apps/coqbot-app/advanced

ejgallego:
I find the command name quite misleading; this won't test the native compiler but the performance of the OCaml compiler.

Zimmi48:
If the version of the OCaml compiler is the same in both OLD and NEW how can you claim that this tests the performance of the OCaml compiler?

Zimmi48:
I guess to test what users will experience. But I also guess this discussion will continue synchronously since you've added a point about it to the next Coq Call.

ejgallego:
> I guess to test what users will experience. But I also guess this discussion will continue synchronously since you've added a point about it to the next Coq Call.

Users will experience this only at install time, so what are we benching here?

Typically we bench vs two Coq commits, so benching that for Coq native we will be measuring the time the OCaml compiler will take to compile the native files, which is fully orthogonal to actual Coq's performance.

ejgallego:
> But is the "installation process" very different from testing performance of native compilation used interactively?

Totally different, the installation process doesn't run the compiled code, which is the point of native compute.

PR #244, opened 2022-10-17
thunk Config.github_private_key

Alizter:
This makes it only be called when needed. Otherwise it gets called during linking and that affects unreleated things.


Zimmi48 (approved):
Happy to merge this, but I would appreciate some more explanation, context or reference to understand better what the issue is.

PR #245, opened 2022-10-17
Add ppx_expect as a dependency

Alizter:
This lets us write inline expectation tests which is quite useful especially in a hard to test application like coqbot.

Depends on #244.

### DIFF
--- a/coq-bot.opam
+++ b/coq-bot.opam
@@ -20,6 +20,7 @@ depends: [
   "yojson" {>= "1.7.0"}
   "bot-components" {dev}
   "toml" {>= "7.0.0"}
+  "ppx_expect" {>= "v0.15.1"}
   "odoc" {>= "1.5.2" & with-doc}
 ]
 build: [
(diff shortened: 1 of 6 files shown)


Zimmi48:
You read my mind!

Zimmi48 (commented):
Can you add an example ppx_expect test somewhere appropriate? Some example I had in mind was in bot-components, when we generate JSON, like in the play_job function.

Alizter:
@Zimmi48 I have a PR coming up with an expectation test. I've implemented the list of failed jobs for the bench. For ppx_expect to work we need to also do #244 as searching for the env var during the test causes an exception (since GITHUB_PRIVATE_KEY is not set).

Alizter:
Let's merge this then I can show the other patch with a test.

Alizter:
@Zimmi48 I can reformat it quickly.

Alizter:
`dune fmt` done, merging.

PR #247, opened 2022-10-18
Extract GitHub endpoint handling.

Zimmi48:
The goal would be to eventually be more declarative in order to add some tests of the results without executing any request.


Zimmi48:
Superseded by the recent refactorings.

Issue #183, opened 2021-11-30
What version of Coq should non-CI minimization default to?


JasonGross:
Currently it's "latest released", and there's no good way to configure it when invoking the minimizer.  Perhaps it should be configurable?

JasonGross:
Maybe something like coqbot minimize optional-coq-version?  This would help with https://github.com/coq/coq/issues/15325#issuecomment-991105413

proux01:
Indeed, note that for coq/coq#15325 I already did the minimization by end, so no hurry.

erikmd:
> Perhaps it should be configurable?

I guess it would be a nice feature indeed!

> Maybe something like coqbot minimize optional-coq-version?

yes… in this case, I guess one could have `optional-coq-version` = [any available tag from the `coqorg` registry](https://hub.docker.com/r/coqorg/coq#supported-tags) ?
e.g.
```
coqbot minimize dev

or

coqbot minimize 8.14-native-ocaml-4.07
```

Issue #250, opened 2022-11-02
"bench native" should be "bench native=on"


SkySkimmer:
As discussed in https://github.com/coq/coq/wiki/Coq-Call-2022-11-02

ejgallego:
Even maybe we could call that `native_cmxs_gen=on`?

But indeed, the problem of the number being quite meaningless in terms of performance still exists, IMHO we need to rethink the bench setup for this to be any useful.

PR #252, opened 2022-11-17
Allow customization of Coq version

JasonGross:
Not entirely clear what the protocol should be, see also #183.  We
currently allow things like `coq-8.14` and `coq.dev` and `coq: dev` and
`ocaml: 4.11`, etc.

### DIFF
--- a/src/bot.ml
+++ b/src/bot.ml
@@ -57,10 +57,13 @@ let callback _conn req body =
     if
       string_match
         ~regexp:
-          ( f "@%s:? [Mm]inimize[^`]*```[^\n]*\n\\(\\(.\\|\n\\)+\\)"
+          ( f "@%s:? [Mm]inimize\\([^`]*\\)```[^\n]*\n\\(\\(.\\|\n\\)+\\)"
           @@ Str.quote bot_name )
         body
-    then Some (Str.matched_group 1 body |> extract_minimize_file)
+    then
+      Some
+        ( Str.matched_group 1 body
+        , Str.matched_group 2 body |> extract_minimize_file )
     else None
   in
   let strip_quoted_bot_name body =
@@ -184,7 +187,7 @@ let callback _conn req body =
       | Ok (_, IssueOpened ({body= Some body} as issue_info)) -> (
           let body = body |> Helpers.trim_comments |> strip_quoted_bot_name in
           match coqbot_minimize_text_of_body body with
-          | Some script ->
+          | Some (options, script) ->
               (fun () ->
                 init_git_bare_repository ~bot_info
                 >>= fun () ->
@@ -192,8 +195,8 @@ let callback _conn req body =
                   ~owner:issue_info.issue.owner ~repo:issue_info.issue.repo
                   (run_coq_minimizer ~script ~comment_thread_id:issue_info.id
                      ~comment_author:issue_info.user
-                     ~owner:issue_info.issue.owner ~repo:issue_info.issue.repo )
-                )
+                     ~owner:issue_info.issue.owner ~repo:issue_info.issue.repo
+                     ~options ) )
               |> Lwt.async ;
               Server.respond_string ~status:`OK ~body:"Handling minimization."
                 ()
@@ -206,7 +209,7 @@ let callback _conn req body =
             comment_info.body |> Helpers.trim_comments |> strip_quoted_bot_name
           in
           match coqbot_minimize_text_of_body body with
-          | Some script ->
+          | Some (options, script) ->
               (fun () ->
                 init_git_bare_repository ~bot_info
                 >>= fun () ->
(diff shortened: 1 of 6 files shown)


JasonGross:
This is deployed and worked correctly at https://github.com/coq/coq/issues/16834#issuecomment-1319231968

@Zimmi48 thoughts on merging?

JasonGross:
This apparently results in a 500 "Error: Internal Server Error" with no log info printed on comments like the one below.  @Zimmi48 Can you help debug?  (I've undeployed this for now)

@coqbot minimize coq.dev
```
true | coqtop
opam install -y coq-fiat-crypto.dev
```

Zimmi48:
OK, I will have a look today.

Zimmi48:
Something happened that made the bot crash. The log says: Process exited with status 143.
Have you observed this pattern of crashing repeatedly?
It's a very surprising behavior, maybe an OOM issue?

JasonGross:
I have not observed it (and did not see such messages in the log), but let me try splitting up the commit into the OCaml part and the bash part, maybe that will reveal what's going on more clearly.  (Though if this only happened about 5 minutes ago, that might be a race condition from me merging #251 or #253 and a redeploy happening?)

JasonGross:
On the other hand I currently see
image
when I click on logentries

JasonGross:
I've split this PR into two commits, one for just the bash change and one for the OCaml change, I'm now trying the deploy of just the bash one

JasonGross:
Bash-only commit seems to be working at https://github.com/coq-community/run-coq-bug-minimizer/issues/2#issuecomment-1320373904

JasonGross:
I see the 143 error only on redeploy.  I continue to see `Error: Internal Server Error` to the minimize comments while also seeing normal things like `Unhandled comment: coqbot merge now`

src/bot.ml
```diff
@@ -57,10 +57,13 @@ let callback _conn req body =
...
         ~regexp:
-          ( f "@%s:? [Mm]inimize[^`]*```[^\n]*\n\\(\\(.\\|\n\\)+\\)"
+          ( f "@%s:? [Mm]inimize\\([^`]*\\)```[^\n]*\n\\(\\(.\\|\n\\)+\\)"
           @@ Str.quote bot_name )
         body
-    then Some (Str.matched_group 1 body |> extract_minimize_file)
+    then
+      Some
+        ( Str.matched_group 1 body
+        , Str.matched_group 2 body |> extract_minimize_file )
```
JasonGross, line 66: The issue seems to be this pair of lines.  Maybe `extract_minimize_file` messes things up in a way that `Str.matched_group 1 body` fails? (Does OCaml evaluate tuples right-to-left?)

SkySkimmer: tuple evaulation order is undefined afaik

Alizter:
Try writing an expect test to test out your regex.

Zimmi48:
> On the other hand I currently see image when I click on logentries

I've seen this once but it worked when I immediately retried.

JasonGross:
Seems to be working now at https://github.com/coq-community/run-coq-bug-minimizer/actions/runs/3500541795/jobs/5863316068 / https://github.com/coq-community/run-coq-bug-minimizer/issues/2#issuecomment-1320576069

JasonGross:
Seems to be working pretty well.  Thoughts on merging @Zimmi48 ?

Zimmi48:
I'm not in a state where I can properly review the PR but if you think this is sufficiently tested (including the case with no argument), then I'm fine with merging. I would just note that it looks like we are starting to grow a language of commands to coqbot and it would be preferable if we try to design this language consistently. It was decided in a Coq Call (then reported as an issue here #250, but not yet implemented) that `@coqbot bench native` should become `@coqbot bench native=on`. This suggests that this command could take options as `@coqbot minimize coq=... ocaml=...`. WDYT?

JasonGross:
> It was decided in a Coq Call (then reported as an issue here #250, but not yet implemented) that `@coqbot bench native` should become `@coqbot bench native=on`. This suggests that this command could take options as `@coqbot minimize coq=... ocaml=...`. WDYT?

I've added support for `coq=...` and `ocaml=...`, and allowed users to specify versions as `8.16`, also `v8.16`, also `V8.16` (they are treated identically).  I didn't feel like specifically forbidding `vdev` and `Vdev`, so these also work.  I've left in other ways of specifying things, like `coq:8.16` and `coq: 8.16` and `coq.8.16` and `coq-8.16`.  I figure it's better to be generous in parsing the user's intent here, even if we only advertise one way of doing things.

JasonGross:
Seems to work both with and without versions at https://github.com/coq-community/run-coq-bug-minimizer/issues/2#issuecomment-1322509333, I'll merge

PR #257, opened 2022-11-24
Improve minimization logic around scripts and files

### LINKED ISSUE #256: coqbot minimize should be able to minimize .v files
`@coqbot minimize` treats its argument as a bash file.  This is fully general, bit it would be nice to support Coq files for convenience.  Maybe we should default to Coq file, and only treat it as bash if the user writes ` ```bash ` or ` ```sh `?

JasonGross:
We now assume that `@coqbot minimize` is followed by a Coq file, not a bash script, unless the user begins with ` ```sh` or ` ```bash` or the script begins with `#!`.

Additionally, we also allow the user to say `@coqbot minimize [text](url)`, and use the `handle-web-file.sh` script in the run-coq-bug-minimizer to automatically download `url` and automagically figure out how to build run coq on it.  (Currently there's support for uncompressing archives, running single .v files, renaming text files to .v if there's no archive, finding _CoqProject, and using autogen, configure, and make to build archives if there's a single makefile.)

Fixes #256

### DIFF
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -1868,6 +1868,10 @@ let pipeline_action ~bot_info pipeline_info ~gitlab_mapping : unit Lwt.t =
               | _ ->
                   Lwt.return_unit ) ) )
 
+type coqbot_minimize_script_data =
+  | MinimizeScript of {quote_kind: string; body: string}
+  | MinimizeAttachment of {description: string; url: string}
+
 let run_coq_minimizer ~bot_info ~script ~comment_thread_id ~comment_author
     ~owner ~repo ~options =
   let options =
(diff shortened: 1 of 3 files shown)


JasonGross:
Seems to work at https://github.com/coq-community/run-coq-bug-minimizer/issues/23#issuecomment-1328097785

Alizter (approved):

Alizter:
@Zimmi48 I had a look at it and am happy. Feel free to merge.

Zimmi48:
Thanks a lot, Ali and Jason! It is great relief for me to see that progress can be made on coqbot, even when I'm less available for reviewing :smile:

PR #262, opened 2023-01-23
Add support for `@coqbot inline-stdlib=yes resume ci minimize ci-foo`

JasonGross:
Also for `@coqbot extra-arg=--inline-coqlib resume ci minimize ci-foo` and `@coqbot extra-arg=--inline-coqlib ci minimize ci-foo`

On top of #257 to avoid conflicts

For https://github.com/coq/coq/pull/16967#issuecomment-1372540099, in conjunction with https://github.com/JasonGross/coq-tools/pull/144

### DIFF
--- a/coq_bug_minimizer.sh
+++ b/coq_bug_minimizer.sh
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 
-# usage: coq_bug_minimizer.sh 'script' comment_thread_id comment_author github_token bot_name bot_domain owner repo coq_version ocaml_version
+# usage: coq_bug_minimizer.sh 'script' comment_thread_id comment_author github_token bot_name bot_domain owner repo coq_version ocaml_version minimizer_extra_arguments
 
 set -ex
 
-if [ $# != 10 ]; then >&2 echo Bad argument count; exit 1; fi
+if [ $# != 11 ]; then >&2 echo Bad argument count; exit 1; fi
 
 script=$1
 comment_thread_id=$2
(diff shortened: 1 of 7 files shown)


JasonGross:
Seems to fail with `Error: Unhandled exception: (Invalid_argument Str.matched_group)`, not sure what's going on

src/helpers.ml
```diff
@@ -2,12 +2,24 @@ open Base
...
   with Stdlib.Not_found -> false
 
+let rec fold_string_matches ~regexp ~f ~init ?(pos = 0) string =
+  if string_match ~regexp ~pos string then
+    let pos = Str.match_end () in
+    f (fun () -> fold_string_matches ~regexp ~f ~init ~pos string)
+  else init
+
+let map_string_matches ~regexp ~f string =
+  fold_string_matches ~regexp ~f:(fun rest -> f () :: rest ()) ~init:[] string
```
SkySkimmer, line 18: execution order is not specified and usually ends up right-to-left so this is wrong
instead do
```suggestion
  fold_string_matches ~regexp ~f:(fun rest -> let v = f () in v :: rest ()) ~init:[] string
```

JasonGross:
This seems to be working [here](https://github.com/coq/coq/pull/16977#issuecomment-1402868303) (well, the coqbot side of it at least; the Coq CI artifact side still has issues at https://github.com/coq/coq/issues/17159).  Shall we merge?

Alizter (approved):

Alizter:
LGTM merge as you wish

PR #263, opened 2023-01-23
Add some string matching utils

JasonGross:
Splitting this out from #262 for ease of review; please suggest more maintainable / idomatic ways of implementing these functions if there are such

### DIFF
--- a/src/helpers.ml
+++ b/src/helpers.ml
@@ -15,7 +15,7 @@ let rec fold_string_matches ~regexp ~f ~init ?(pos = 0) string =
   else init
 
 let map_string_matches ~regexp ~f string =
-  fold_string_matches ~regexp ~f:(fun rest -> f () :: rest ()) ~init:[] string
+  fold_string_matches ~regexp ~f:(fun rest -> let v = f () in v :: rest ()) ~init:[] string
 
 let iter_string_matches ~regexp ~f string =
   fold_string_matches ~regexp ~f:(fun rest -> f () ; rest ()) ~init:() string
--- a/src/helpers.ml
+++ b/src/helpers.ml
@@ -2,11 +2,26 @@ open Base
 
 let f = Printf.sprintf
 
+let rec fold_string_matches ~regexp ~f ~init ?(pos = 0) string =
+  ( try
+      let _ = Str.search_forward (Str.regexp regexp) string pos in
+      true
+    with Stdlib.Not_found -> false )
+  |> function
+  | true ->
+      let pos = Str.match_end () in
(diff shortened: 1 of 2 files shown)


JasonGross:
Note, btw, that really all I need is something equivalent to python's [`re.findall`](https://docs.python.org/3/library/re.html#re.findall), and I'm happy with whatever is the most sensible way of implementing it

Alizter:
I've always thought we should use a library like https://github.com/ocaml/ocaml-re which is much easier to use, without any global state mutation.

src/helpers.ml
```diff
@@ -2,11 +2,26 @@ open Base
 
 let f = Printf.sprintf
 
+let rec fold_string_matches ~regexp ~f ~init ?(pos = 0) string =
+  ( try
+      let _ = Str.search_forward (Str.regexp regexp) string pos in
```
Alizter, line 7:
```suggestion
      let () = Str.search_forward (Str.regexp regexp) string pos in
```
I think it is good practice not to leave the pattern empty to make sure it is really evaluated.

Alizter (approved):

src/helpers.ml
JasonGross: `search_forward` returns an `int`, not a unit.  Also the previous code was already
```
try
    let _ = Str.search_forward (Str.regexp regexp) string 0 in
    true
  with Stdlib.Not_found -> false
```

Alizter: In that case could you do `(_ : int)`.

JasonGross: Would
```ocaml
let string_match ~regexp ?(pos = 0) string =
  try
    let _ = Str.search_forward (Str.regexp regexp) string pos in
    true
  with Stdlib.Not_found -> false

let rec fold_string_matches ~regexp ~f ~init ?(pos = 0) string =
  if string_match ~regexp ~pos string then
    let pos = Str.match_end () in
    f (fun () -> fold_string_matches ~regexp ~f ~init ~pos string)
  else
      init
```
be clearer?

Alizter: That would be clearer but my suggestion was to make the code a bit more rigid to changes. If an argument is omitted during refactoring say, then suddenly the search_forward won't work. By adding `(_ : int)` you guarantee that it will have been evaluated.

Not super important anyway, merge if you wish.

Zimmi48:
FTR, I wouldn't mind introducing a new library if it helps. But we should probably add more expect tests so that we know that we are not breaking anything (because errors in regexp are typically not caught by typing and a frequent cause of runtime failures when we introduce new ones, so I wouldn't be surprised if it caused havoc if we adopt a new regexp library).

PR #274, opened 2023-04-03
Strip spaces from quote_kind

JasonGross:
This should fix the issue where ` ```bash ` is not recognized because there's apparently a trailing space or `\r`.


Zimmi48 (approved):
This was test-deployed two weeks ago. I assume this has been tested. If that's the case, feel free to go ahead and merge.

Issue #234, opened 2022-06-27
Make coqbot workaround the github-to-gitlab 20' lag for syncing the main branches?


erikmd:
### Context

I remember discussing this issue a long time ago with @Zimmi48 and it seemed "minor" (and/or not essential for maintaining coq/coq), but it actually appeared blocking to me since this comment: https://github.com/coq-community/docker-coq/pull/50#issuecomment-1165749733

Cc @Alizter @JasonGross also FYI

### Current behavior

* coqbot is extremely nice for syncing GitHub PR branches → GitLab branches named `pr-*`

* However when a repository such as https://github.com/coq-community/docker-coq **strongly relies on continuous deployment at merge time in the default branch**, it is very annoying to face with this "documented sync lag" [`pull_mirror_interval_seconds`](https://docs.gitlab.com/ee/administration/instance_limits.html#pull-mirroring-interval).

* Namely, the repository is automatically synced only **every 30'** (it is easy to notice this by monitoring the text `Successfully updated N minutes ago` at URL https://gitlab.com/coq-community/docker-coq/-/branches)

* Furthermore, a manual sync is only possible **every 5'** as documented here:
  https://docs.gitlab.com/ee/user/project/repository/mirror/#update-a-mirror

### Expected behavior

It would be nice if coqbot can automatically push the default branch (e.g. `master` or `main` or *a set of user-specified branches*) to GitLab as soon as it detects a push to this/these branches in GitHub.

As a result, with this feature, these two PRs could be merged quasi-simultaneously in this order:
1. https://github.com/coq-community/docker-coq/pull/50
2. https://github.com/coq/coq/pull/15560

### Misc notes

Regarding the potential "risk" that should be tested:

* If coqbot can push master from github to gilab before the automatic mirroring, one should check that this manual push won't prevent the upcoming, automatic mirroring of (all other) branches.
* but I guess this shouldn't occur if coqbot only do fast-forward pushes (unlike `pr-*` pushes that are typically done with `--force`)

Zimmi48:
What about entirely disabling the mirroring for projects which choose to use coqbot for this task?

erikmd:
@Zimmi48

> What about entirely disabling the mirroring for projects which choose to use coqbot for this task?

Yes of course, if coqbot could perform the "fast-forward push mirroring of all branches" itself, why not indeed!

Zimmi48:
[...]

PR #276, opened 2023-04-17
Mirror feature + multi-GitLab instance support.

Zimmi48:
This is the result of work with @erikmd over the last month or so.

The first commit has been tested and adds support for mirroring a hard-coded selection of GitHub repositories to GitLab (we plan to make the feature more largely available through configuration files, which would complete #234, see the message for the first commit).

The second commit has not been tested yet and adds support for multiple GitLab instances. It requires supporting multiple bot names as not all GitLab instances will allow creating a new user with the same username. (The name of the bot on gitlab.inria.fr will be `x-CBot`).

### DIFF
--- a/src/helpers.ml
+++ b/src/helpers.ml
@@ -75,44 +75,39 @@ let github_repo_of_gitlab_project_path ~gitlab_mapping ~gitlab_domain
 
 let parse_gitlab_repo_url ~http_repo_url =
   if not (string_match ~regexp:"https?://\\([^/]*\\)/\\(.*/.*\\)" http_repo_url)
-  then failwith "Could not match project name on repository url.\n" ;
-  (Str.matched_group 1 http_repo_url, Str.matched_group 2 http_repo_url)
+  then Error (f "Could not parse GitLab repository URL %s." http_repo_url)
+  else Ok (Str.matched_group 1 http_repo_url, Str.matched_group 2 http_repo_url)
+
+let parse_gitlab_repo_url_and_print ~http_repo_url =
+  match parse_gitlab_repo_url ~http_repo_url with
+  | Ok (gitlab_domain, gitlab_repo_full_name) ->
+      Stdio.printf "GitLab domain: \"%s\"\n" gitlab_domain ;
+      Stdio.printf "GitLab repository full name: \"%s\"\n" gitlab_repo_full_name
+  | Error msg ->
+      Stdio.print_endline msg
 
 let%expect_test "http_repo_url_parsing_coq" =
-  let gitlab_domain, gitlab_repo_full_name =
-    parse_gitlab_repo_url ~http_repo_url:"https://gitlab.com/coq/coq"
-  in
-  Stdio.print_endline gitlab_domain ;
-  Stdio.print_endline gitlab_repo_full_name ;
+  parse_gitlab_repo_url_and_print ~http_repo_url:"https://gitlab.com/coq/coq" ;
   [%expect {|
-     gitlab.com
-     coq/coq |}]
+     GitLab domain: "gitlab.com"
+     GitLab repository full name: "coq/coq" |}]
 
 let%expect_test "http_repo_url_parsing_mathcomp" =
-  let gitlab_domain, gitlab_repo_full_name =
-    parse_gitlab_repo_url
-      ~http_repo_url:"https://gitlab.inria.fr/math-comp/math-comp"
-  in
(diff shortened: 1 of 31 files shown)


Zimmi48 (commented):
State of my partial review (first commit + not even halfway through the second commit) before meeting with Erik today.

src/helpers.ml
```diff
@@ -50,28 +50,67 @@ let trim_comments comment =
...
+
+let%expect_test "http_repo_url_parsing_coq" =
+  let gitlab_domain, gitlab_repo_full_name =
+    parse_gitlab_repo_url ~http_repo_url:"https://gitlab.com/coq/coq"
+  in
+  Stdio.print_endline gitlab_domain ;
+  Stdio.print_endline gitlab_repo_full_name ;
+  [%expect {|
+     gitlab.com
+     coq/coq |}]
```
erikmd, line 87: @Zimmi48 do you confirm that:

* the `let%expect_test … [%expect {|…|}]` perform some string trimming automatically?
  * if yes, maybe this test is not strict enough? (e.g., we wouldn't want to get `"coq/coq "` with a trailing space)
  * if not, I believe the test will fail?
* what is the command-line necessary to run all the tests? (admittedly, was lazy to double-check the doc :)
* I guess these new tests are not already run in a GHA CI test:
  * do you think it would be practical to add a GHA workflow within this PR, or after merging it? or before?
  * in the later case, may I suggest to use `coq-community/docker-coq-action@v1`… because it is definitely not Coq-specific, but opam-specific :)

Zimmi48: >the `let%expect_test … [%expect {|…|}]` perform some string trimming automatically?

Yes, on the expected output. I don't know if it trims the actual output.

>if yes, maybe this test is not strict enough? (e.g., we wouldn't want to get `"coq/coq "` with a trailing space)

Possibly. This should be tested.

>if not, I believe the test will fail?

Currently, it does not fail.

>what is the command-line necessary to run all the tests? (admittedly, was lazy to double-check the doc :)

`dune runtest` IIRC

>I guess these new tests are not already run in a GHA CI test:
>       
>* do you think it would be practical to add a GHA workflow within this PR, or after merging it? or before?
>* in the later case, may I suggest to use `coq-community/docker-coq-action@v1`… because it is definitely not Coq-specific, but opam-specific :)

Sure, we can add a GH Action. Currently, there is none that runs on PRs. Feel free to propose a setup in a separate PR.

Zimmi48:
The current version of the PR still has issues. Retrying a job led to the following error:

```
Error when retrying job projects/6138686/jobs/4191368974: I don't know about GitLab domain https://gitlab.com/coq/coq (not in my configuration file).
```

The problem comes from here: https://github.com/coq/bot/blob/8f9c8b982baa0e22911f8642ae32b2799bcb1600/src/bot.ml#L481-L483

The first part of the `external_id` is not the GitLab domain but the full repo URL. It should be parsed to get the GitLab domain.

Zimmi48:
Rebased and redeploying on top of current master. We should probably consider merging.

erikmd:
Hi @Zimmi48, thanks a lot for the bugfix and the polishing!

To sum up my viewpoint on the PR:

* The features indeed look production ready to me 👍
* But I'm not 100% sure yet that the added tests are optimal w.r.t. their form (but admittedly, this can be polished in a later PR)
* Finally if you opt to merge it now, no objection, but mind to temporarily disable the `math-comp/math-comp` hardcoded binding for gitlab inria, given it is not ready yet w.r.t. the mathcomp repo (sorry, I've got 0 free time this week;
I'll let you know when it's ready)

Zimmi48:
> * But I'm not 100% sure yet that the added tests are optimal w.r.t. their form (but admittedly, this can be polished in a later PR)

I think this is fixed now.

> * Finally if you opt to merge it now, no objection, but mind to temporarily disable the `math-comp/math-comp` hardcoded binding for gitlab inria, given it is not ready yet w.r.t. the mathcomp repo (sorry, I've got 0 free time this week;
>   I'll let you know when it's ready)

The PR contains a last commit disabling the math-comp/math-comp configuration. But we could also drop it before merging if you think that math-comp/math-comp will be ready to use this feature soon.

Zimmi48:
@erikmd I'm going to merge this PR (including the last commit disabling it on math-comp/math-comp) because it is inconvenient having this PR being the deployed version of coqbot for too long (it makes other changes more complicated).

PR #297, opened 2024-01-11
Strip out more ANSI color codes

JasonGross:
This should issues like
https://github.com/coq/coq/pull/18446/checks?check_run_id=20043646334 not finding `Error` since the introduction of colors in https://github.com/coq/coq/pull/18338.  ANSI color codes sometimes include three numeric components.  If we want to be more precise, we could use `([0-9]+;?){1,3}` or even `([0-9]+;){0,2}[0-9]+`, but I think this might be overkill, and there's not too much harm in using the simpler expression.

### DIFF
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -25,7 +25,7 @@ let send_status_check ~bot_info job_info ~pr_num (gh_owner, gh_repo)
   in
   let trace_lines =
     trace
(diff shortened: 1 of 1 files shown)


JasonGross:
This seems to work, so I'm going to merge so it's not lost.  @Zimmi48 or others, feel free to rework with a different regex if desired.

PR #305, opened 2024-06-17
Migrate release management to GitHub projects v2

### LINKED ISSUE #278: Migrating the backporting tracking mechanism to Projects V2.
In https://github.com/orgs/coq/projects/8 (private and only visible to core devs and bot maintainers for now, to avoid polluting most users with this experiment), I have experimented with a backporting tracking project using the new GitHub Projects.

One advantage that I can see is that instead of creating a new project for each new release branch, we could use a single project with views per release branches (I have separated one view for "Request inclusion in ..." and one view for "Shipped in ...").

The idea of the workflow would be the following:
- When a pull request is merged, coqbot looks at its milestone. If the milestone's description contains a request to backport to one (or several) release branch(es), then coqbot will:
  1. Add the pull request to this project (always the same) and retrieve the item ID.
  2. Search for existing single-select fields in the project named "vX.X status" for each VX.X branch to which backporting was requested.
  3. Retrieve the fields' ID and the IDs of their values if they exist, create them if they don't exist yet (always with the same two values "Request inclusion" and "Shipped", with the same colors and the same descriptions).
  4. In the case where the fields are created, also create new views in the project.
  5. Set the value of these fields to "Request inclusion" or "Shipped" for the newly added item depending on whether the pull request was merged in `master` or in the release branch.
- When a pull request is backported to a release branch, coqbot looks for the field in the project corresponding to the release branch and sets the value of this field to "Shipped".

Currently, there is a limit at 1200 non-archived items in a GitHub project, which means that from time to time we might need to archive items corresponding to old releases and delete the corresponding views.

### LINKED ISSUE #303: Deprecation in GraphQL API to address by the end of June.
After updating the GitHub GraphQL schema in e5e0353, building yields the following deprecation message:

Warning 22 [preprocessor]: Field "databaseId" has been deprecated. Reason: `databaseId` will be removed because it does not support 64-bit signed integer identifiers. Use `fullDatabaseId` instead. Removal on 2024-07-01 UTC.

We need to move from `databaseId` to `fullDatabaseId` on this line (before end of June):

https://github.com/coq/bot/blob/e0942d05267a66cd146e8218462ecd6c020d823f/bot-components/GitHub_GraphQL.ml#L55

However, `databaseId` was of type `Int` while `fullDatabaseId` is of type `BigInt`. This is thus no longer automatically parsed as an int. We should investigate how to fix the code to allow this migration.

Any help is welcome!

Zimmi48:
Partially implements #278 (~~without removing the project V1 support, so that we can test this while keeping the other one for now~~). What is missing:

- [x] Automatically create field if it does not exist yet.
- [ ] Create views when creating a new field.
- [x] Management of PRs removed from the project (now will use a rejected field value instead).
- [ ] Update https://github.com/coq/coq/blob/master/dev/doc/release-process.md.

@proux01 @silene I intend to test-deploy this, so https://github.com/orgs/coq/projects/11/views/4 and https://github.com/coq/coq/projects/42 should in principle be updated concurrently when new PRs are merged or backported in the 8.20+rc1 milestone.

Close coq/coq#19156.
Close #278.
Close #303.

### DIFF
--- a/bot-components/GitHub_GraphQL.ml
+++ b/bot-components/GitHub_GraphQL.ml
@@ -2,32 +2,12 @@
 
 (* Queries *)
 
-module PullRequest_Milestone_and_Cards =
+module PullRequest_Cards =
 [%graphql
 {|
-  fragment Column on ProjectColumn {
-    id
-    databaseId
-  }
-
-  query backportInfo($owner: String!, $repo: String!, $number: Int!) {
+  query prCards($owner: String!, $repo: String!, $number: Int!) {
     repository(owner: $owner,name: $repo) {
       pullRequest(number: $number) {
-        milestone {
-          title
-          description
-        }
-        projectCards(first:100) {
-          cards: nodes {
-            card_id: id
-            column { ... Column }
-            project {
-              columns(first:100) {
-                nodes { ... Column }
-              }
-            }
-          }
-        }
         projectItems(first: 100) {
           items: nodes {
             item_id: id
@@ -60,7 +40,6 @@ module PullRequest_ID_and_Milestone =
     repository(owner: $owner,name: $repo) {
       pullRequest(number: $number) {
         id
-        databaseId
         milestone {
           title
           description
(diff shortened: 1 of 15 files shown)


Zimmi48:
I have now fixed the issues I discovered this morning. I will merge this pull request to avoid staying longer with a deployment which is not from the main branch. I will address the remaining todos (documentation and auto-creating views) later.

PR #311, opened 2024-09-16
Simplify Docker image build.

Zimmi48:
In principle, the second command is not needed since coq-bot already depends on bot-components. Furthermore, since a few days, this second command seems to consistently trigger a timeout in our CI during the Docker build. So removing it could help fix this issue.


Zimmi48:
Tests on my fork confirms that this fixes the deployment issue.

PR #316, opened 2024-10-12
`dune build @fmt --auto-promote` with ocamlformat 0.26.2 

JasonGross:
Alternative to #315 
Closes #315


Zimmi48 (approved):
That's the version I have in `nix-shell` so that's the one we should use. Thanks!

PR #319, opened 2024-10-13
Post a comment when the bug minimizer fails to start, even when not in CI mode

JasonGross:


JasonGross:
I'm going to merge this, since I don't think it can hurt (and I've tested that it deploys fine)

PR #320, opened 2024-10-13
Include stderr in error reports for easier debugging

### LINKED ISSUE #318: execute_cmd should include process stderr when the exit code is not 0
The message in https://github.com/coq/coq/pull/19680#issuecomment-2408603227 would be more useful if https://github.com/coq/bot/blob/7e8b31fd51e1482757bdf1db1d28e8c2c2049a1f/src/git_utils.ml#L85-L97 included stderr

JasonGross:
On top of #317
Fixes #318

<strike>Except I think it doesn't quite work in the current form</strike>

### DIFF
--- a/src/git_utils.ml
+++ b/src/git_utils.ml
@@ -11,10 +11,13 @@ let gitlab_repo ~bot_info ~gitlab_domain ~gitlab_full_name =
   |> Result.map ~f:(fun token ->
          f "https://oauth2:%s@%s/%s.git" token gitlab_domain gitlab_full_name )
 
-let report_status ?(mask = []) command report code =
+let report_status ?(mask = []) ?(stderr_content = "") command report code =
+  let stderr =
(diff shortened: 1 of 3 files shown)


JasonGross:
@Zimmi48 Do you have any opinion on the implementation here?  I've checked that it still emits stderr and stdout in the log, doesn't crash, etc.

Zimmi48 (approved):
I can't say that I understand precisely all the code (in particular the `copy_stream` function), but this looks good to me.

JasonGross:
> I can't say that I understand precisely all the code (in particular the `copy_stream` function), but this looks good to me.

`copy_stream` reads from a steam character by character, simultaneously saving it to a buffer and writing it to another stream.  It was written mostly by claude.ai.  It's possibly a bit overkill (we could just read from stdout and stderr all at once), but this version keeps them interleaved in the logs.

PR #322, opened 2024-10-13
Longer timeout for depext

JasonGross:
It seems to require about 70 seconds, let's give it 5 minutes in case it gets even slower in the future.


JasonGross:
I'm going to merge this, since it works on my end

PR #330, opened 2024-11-14
Adapt to coq/coq#19530.

Zimmi48:


Zimmi48:
Test deployed

Zimmi48:
It worked:

image

One limitation is that the links for the refman for `doc:ci-refman` and `doc:refman` are hard to distinguish. It could be solved by not including the `doc:refman` target in the full CI (only for the light CI).

PR #337, opened 2025-03-24
Fix auto minimizer parsing the job summary

SkySkimmer:


SkySkimmer:
You can make a PR with that if you want

JasonGross:
Let's try it this way first

JasonGross (approved):

PR #344, opened 2025-07-04
Include option to disable minimization suggestions in PR body.

### LINKED ISSUE #343: Minimizer advertisement should be opt-in
We could add a comment like
```
<!--
# Turn this on if you want coq-bot remind you commands to run the minimizer
# (you can always call the minimizer with @coqbot minimize anyway)
offer-minimizer: off
-->
```
in the PR template. Coqbot would look at it and only send the message advertizing the minimizer when it finds the string `offer-minimizer: on`.
That would greatly reduce the amount of unsolicited messages for developers.

Cc @Zimmi48 following a discussion last week.

Zimmi48:
Closes #343.

I'm going to test-deploy this. @proux01 can you craft a PR with `offer-minimizer: off` in the body and a code change that would normally make the minimizer offer suggestions to test that it works?

EDIT: to check the deployment status: https://github.com/Zimmi48/bot/actions/runs/16068760889

### DIFF
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -1021,6 +1021,7 @@ type ci_minimization_pr_info =
   ; head: string
   ; pr_number: int
   ; draft: bool
+  ; body: string
   ; labels: string list
   ; base_pipeline_finished: bool
   ; head_pipeline_finished: bool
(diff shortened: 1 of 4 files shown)


src/actions.ml
```diff
@@ -1298,6 +1300,10 @@ let suggest_ci_minimization_for_pr = function
       RunAutomatically
   | {labels} when List.exists ~f:(String.equal "kind: infrastructure") labels ->
       Silent "this PR is labeled with kind: infrastructure"
+  | {body} when String.is_substring ~substring:"offer-minimizer: off" body ->
+      Silent
+        "the PR body contains an 'offer-minimizer: off' directive, which turns \
+         off minimization suggestions"
```
proux01, line 1306: I think the logic should be reverted: only `Suggest` when finding the string `offer-minimizer: on`
It's common for experienced developers to wipe out the template for some trivial PRs and you don't want the spam in that case.

proux01: Of course, this doesn't prevent still having the thing opt-out with the template being `offer-minimizer: on` by default.

Zimmi48:
I will open a PR with the code for inverting the logic for future reference.

PR #345, opened 2025-07-04
Invert the logic of the auto-minimization suggestions.

Zimmi48:
To be merged only after the pull request template has been modified for a few months.


Zimmi48:
I guess so. Let's watch out for confused developers who don't get their minimization suggestions though! :laughing:

PR #348, opened 2025-09-28
Report more detailed information when failing to find pipeline checks

### LINKED ISSUE #347: Bot should report more errors when head check is not found in ci minimization
https://github.com/rocq-prover/rocq/pull/21098#issuecomment-3343594636 is not sufficiently explanatory, we should modify
https://github.com/rocq-prover/bot/blob/8eb0fdaef1c4e5c2ff4243b1bb5dc97d7eb2d70d/src/actions.ml#L1187-L1196
to also check whether one of the error checks is the pipeline check (and print the error message in that case), and otherwise to also list out the error head checks, and similarly for base checks.

JasonGross:
Fix #347

### DIFF
--- a/src/actions.ml
+++ b/src/actions.ml
@@ -1103,21 +1103,31 @@ let fetch_ci_minimization_info ~bot_info ~owner ~repo ~pr_number
                   | None ->
                       `Trd check_tab_info )
           in
+          let extract_pipeline_check_errors =
+            List.filter ~f:(fun (name, _) ->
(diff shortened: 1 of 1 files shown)


Zimmi48:
Sorry, I've merged a PR, and so the test deployment is not current anymore. That being said, this PR was tested successfully at: https://github.com/rocq-prover/rocq/pull/21098#issuecomment-3346520472
It seems to have provided a hint of the reason of the failure: https://github.com/rocq-prover/rocq/pull/21098#issuecomment-3346649994

PR #364, opened 2026-01-22
Fix bug with run full CI in review comment.

Zimmi48:
Fixes the issue observed in https://github.com/rocq-prover/rocq/pull/21416#issuecomment-3785084993 and reported in https://rocq-prover.zulipchat.com/#narrow/channel/243318-coqbot-devs-.26-users/topic/bot.20confused.


Zimmi48:
Apparently, this bug was present from the point when the feature was introduced, and was never hit until now :open_mouth: 

38c93f5a7f2a7f7d08b4c92beafa296e9d4ef444

PR #370, opened 2026-07-29
Fix dev setup: gitignore, formatting, and schema build rules

lykimq:
- Add missing files to `.gitignore` (credentials, generated files, editor artifacts)
- Reformat OCaml source files via `dune build @fmt --auto-promote`
- Change schema fetch rules (`github-schema.json`, `gitlab-schema.json`) from `mode promote` to `mode fallback` so `dune build` no longer triggers network fetches on every run; add named aliases (`update-github-schema`, `update-gitlab-schema`) for explicit schema refresh.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -21,23 +21,25 @@ ocaml libraries.
 Use the following command to build:
 
 ```
-dune build --ignore-promoted-rules
-```
-
-If you want to update the GraphQL schema:
-
-```
 dune build
 ```
 
-This call to `dune build` without the `--ignore-promoted-rules` option
-requires that a file `bot-components/.github-token` be provided and
-contain a single line with a GitHub API personal token (with no
-specific permission).  It will use this token and the node package
-`get-graphql-schema` to update the GitHub schema stored in
-`bot-component/schema.json`, before building the project.
-Get `get-graphql-schema` with `npm install get-graphql-schema -g` if you're
-not compiling in `nix-shell`.
+Committed GraphQL schemas under `bot-components/github/github-schema.json` and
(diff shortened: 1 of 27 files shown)


Zimmi48:
I agree with changing the default to not re-fetch the schema every time, but:

- the doc should be updated (currently it says to use `dune build --ignore-promoted-files` for this reason, there's even one use in CI)
- I tested the alias and it doesn't seem to work to regenerate the files without the universe dependency (unless I am missing an option, which should be documented).

BTW, it would be very convenient to have CI jobs to:
- auto-check formatting so that we don't have to fix it anymore
- open PRs and auto-merge them if CI passes to update the GraphQL schemas (BTW, I just saw that updating the GitHub GraphQL schema breaks the build, but I didn't investigate further yet).

lykimq:
- the doc should be updated (currently it says to use dune build --ignore-promoted-files for this reason, there's even one use in CI) 94e73ac95b87fd69112d74d8d02aff425659f892 and [6026f35](https://github.com/rocq-prover/bot/pull/370/commits/6026f352db55ddb62cf81be728a9ce99b6606389)
- I tested the alias and it doesn't seem to work to regenerate the files without the universe dependency (unless I am missing an option, which should be documented).
I have fixed and update it in the commit 8219a777450b7c40da7a70b0431854bbb75dcfbd

lykimq:
BTW, it would be very convenient to have CI jobs to:

auto-check formatting so that we don't have to fix it anymore
open PRs and auto-merge them if CI passes to update the GraphQL schemas (BTW, I just saw that updating the GitHub GraphQL schema breaks the build, but I didn't investigate further yet).

For those I will open another PRs.

Zimmi48 (approved):

PR #372, opened 2026-07-30
Allow passing configuration options to the `bench` command

Janno:
This *should* allow writing commands like `bench coq_opam_packages="a b c"`, eliminating the need to push temporary commits for bench configuration. The native bench branch in the parser is now gone as that can be written as `bench coq_native` now.

The commits are hopefully self-explanatory. I took the liberty to restructure the comment parsing code (commit #1) because the old code was quite confusing to me and the control flow seemed odd. We now commit to one command and if argument parsing fails or permissions are missing, we go straight to the error case.

The string parsing and the test cases are generated by LLMs. The rest of the code is handwritten.

The PR is a draft for three reasons:
- ~~Passing arbitrary variables to the bench job could be a security concern. I don't know the bench code well enough to say. We might want a whitelist?~~ (EDIT: No change in security risk compared to setting arguments via new commits that change bench.sh)
- I have no idea how to run all tests in the repo. A bunch of dependencies are not recorded (`alcotest`, `ppx_expect`) and adding them manually still brings me to a point where some kind of token is required. Disabling all that still produces failures in inline tests that I do not understand.
- ~~The bench command is entirely undocumented and with it being a bit more complex now I think we need documentation.~~ (EDIT: I found the actual place where the bench command is documented (https://github.com/rocq-prover/rocq/wiki/Benchmarking) and since it is not in this repo, it does not affect the actual code changes.)

### DIFF
--- a/bot-components/utils/String_utils.ml
+++ b/bot-components/utils/String_utils.ml
@@ -48,6 +48,110 @@ let first_line_of_string s =
 let remove_between s i j =
   String.sub ~pos:0 ~len:i s ^ String.sub s ~pos:j ~len:(String.length s - j)
 
+let split_shell_words ~preserve_syntax input =
+  let is_whitespace = function
+    | ' ' | '\t' | '\n' | '\r' | '\011' | '\012' ->
+        true
+    | _ ->
+        false
+  in
+  let buffer = Buffer.create (String.length input) in
+  let add_syntax char = if preserve_syntax then Buffer.add_char buffer char in
+  let finish_token token_started tokens =
+    if token_started then Buffer.contents buffer :: tokens else tokens
+  in
+  let length = String.length input in
+  let rec split index quote token_started tokens =
+    if index = length then
+      match quote with
+      | Some delimiter ->
+          Error (Printf.sprintf "unterminated %c quote" delimiter)
+      | None ->
+          Ok (List.rev (finish_token token_started tokens))
+    else
+      let char = input.[index] in
+      match quote with
+      | None when is_whitespace char ->
+          let tokens = finish_token token_started tokens in
+          Buffer.clear buffer ;
+          split (index + 1) None false tokens
+      | None when Char.equal char '\\' ->
+          if index + 1 = length then Error "trailing escape character"
+          else
+            let escaped = input.[index + 1] in
+            if Char.equal escaped '\n' then
+              split (index + 2) None token_started tokens
+            else (
+              add_syntax char ;
+              Buffer.add_char buffer escaped ;
+              split (index + 2) None true tokens )
+      | None when Char.equal char '\'' || Char.equal char '"' ->
+          add_syntax char ;
+          split (index + 1) (Some char) true tokens
+      | None ->
+          Buffer.add_char buffer char ;
+          split (index + 1) None true tokens
+      | Some '\'' when Char.equal char '\'' ->
+          add_syntax char ;
+          split (index + 1) None true tokens
+      | Some '\'' ->
+          Buffer.add_char buffer char ;
+          split (index + 1) quote true tokens
+      | Some '"' when Char.equal char '"' ->
+          add_syntax char ;
+          split (index + 1) None true tokens
+      | Some '"' when Char.equal char '\\' ->
+          if index + 1 = length then Error "unterminated \" quote"
+          else
+            let escaped = input.[index + 1] in
+            if Char.equal escaped '\n' then
+              split (index + 2) quote true tokens
+            else if
+              List.mem ['"'; '\\'; '$'; '`'] escaped ~equal:Char.equal
+            then (
+              add_syntax char ;
+              Buffer.add_char buffer escaped ;
+              split (index + 2) quote true tokens )
+            else (
+              Buffer.add_char buffer char ;
+              split (index + 1) quote true tokens )
+      | Some '"' ->
+          Buffer.add_char buffer char ;
+          split (index + 1) quote true tokens
+      | Some _ ->
+          failwith "unsupported quote delimiter"
+  in
+  split 0 None false []
+
+let split_on_unquoted_whitespace input =
+  split_shell_words ~preserve_syntax:true input
+
+let parse_key_value_arguments input =
+  match split_shell_words ~preserve_syntax:false input with
+  | Error _ as error ->
+      error
+  | Ok arguments ->
+      let rec parse parsed = function
+        | [] ->
+            Ok (List.rev parsed)
+        | argument :: arguments -> (
+          match Stdlib.String.index_opt argument '=' with
+          | None when String.is_empty argument ->
+              Error (Printf.sprintf "argument %S has an empty key" argument)
+          | None ->
+              parse ((argument, None) :: parsed) arguments
+          | Some 0 ->
+              Error (Printf.sprintf "argument %S has an empty key" argument)
+          | Some separator ->
+              let key = String.sub argument ~pos:0 ~len:separator in
+              let value =
+                String.sub argument ~pos:(separator + 1)
+                  ~len:(String.length argument - separator - 1)
+              in
+              parse ((key, Some value) :: parsed) arguments )
+      in
+      parse [] arguments
+
 (******************************************************************************)
 (* Formatting Functions                                                       *)
 (******************************************************************************)
--- a/bot-components/utils/String_utils.ml
+++ b/bot-components/utils/String_utils.ml
@@ -67,7 +67,9 @@ let split_shell_words ~preserve_syntax input =
     if index = length then
       match quote with
       | Some delimiter ->
-          Error (Printf.sprintf "unterminated %c quote" (match delimiter with Single -> '\'' | Double -> '"'))
+          Error
+            (Printf.sprintf "unterminated %c quote"
+               (match delimiter with Single -> '\'' | Double -> '"') )
       | None ->
           Ok (List.rev (finish_token token_started tokens))
     else
@@ -109,10 +111,8 @@ let split_shell_words ~preserve_syntax input =
           if index + 1 = length then Error "unterminated \" quote"
           else
             let escaped = input.[index + 1] in
-            if Char.equal escaped '\n' then
-              split (index + 2) quote true tokens
-            else if
-              List.mem ['"'; '\\'; '$'; '`'] escaped ~equal:Char.equal
+            if Char.equal escaped '\n' then split (index + 2) quote true tokens
+            else if List.mem ['"'; '\\'; '$'; '`'] escaped ~equal:Char.equal
             then (
               add_syntax char ;
               Buffer.add_char buffer escaped ;
--- a/bot-components/utils/String_utils.ml
+++ b/bot-components/utils/String_utils.ml
@@ -48,6 +48,8 @@ let first_line_of_string s =
 let remove_between s i j =
   String.sub ~pos:0 ~len:i s ^ String.sub s ~pos:j ~len:(String.length s - j)
 
+type quote = Single | Double
+
 let split_shell_words ~preserve_syntax input =
   let is_whitespace = function
     | ' ' | '\t' | '\n' | '\r' | '\011' | '\012' ->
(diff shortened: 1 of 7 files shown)


SkySkimmer:
>   Passing arbitrary variables to the bench job could be a security concern

No more than pushing a PR where the script sets those variables and having the bot run it. That's why the bot bench command is only allowed for authorized people.

Janno:
I am undrafting this. Docs live elsewhere, there is no change in security risk, and judging by the contributing guide, I would need to do some serious work for testing that I do not have time for right now. I wonder if it isn't easier to merge it and revert the merge if anything goes wrong. (Assuming the changes look good to everyone.)

Zimmi48:
The tests should have been fixed in #375. We could merge this other one first if we want to make sure that everything works fine.

Zimmi48:
If you want, I can also do a test deployment of this branch so that you can test it in production before we merge.

src/webhooks/github.ml
```diff
@@ -93,125 +93,158 @@ let handle_comment_created ~bot_info ~key ~app_id ~github_bot_name
...
+         Bot_components.Github_installations.action_as_github_app ~bot_info
+           ~key ~app_id ~owner:comment_info.issue.issue.owner
+           (fun ~bot_info ->
+              Bench.run_bench ~bot_info
+                ~key_value_pairs:[("coq_native", "yes")]
+                comment_info ) )
+      |> Lwt.async ;
+      Server.respond_string ~status:`OK
+        ~body:(f "Received a request to start the bench.")
+        ()
```
lykimq, line 225: This pattern is unreachable because of the function `parse_bench_native` has been deleted. It was the function that produced `BenchNative`.

So we can remove it.

src/webhooks/github.ml
```diff
@@ -93,125 +93,158 @@ let handle_comment_created ~bot_info ~key ~app_id ~github_bot_name
...
+        string_match
+          ~regexp:(f {|@%s:? [Bb]ench\( *$\| +\([^\n]+\) *\(\n\|$\)\)|} @@ Str.quote github_bot_name)
+          body
+      then
+        match Str.matched_group 2 body with
+        | exception _ -> Some (`Bench [])
+        | args ->
+          match parse_key_value_arguments args with
+          | Result.Ok args -> Some (`Bench args)
+          | _ -> None
```
lykimq, line 137: Silent parse failures

When `parse_key_value_arguments` returns `Error`, we map that to `None` and end up as "Unhandled comment". 

Could we return an explicit error command (or post a comment) with the parse error instead of dropping it.

Something like this:
```
| Result.Ok args -> Some (`Bench args)
| Result.Error err -> Some (`BenchParseError err)
```

Then handle it explicitly:
```
Some (`BenchParseError err) ->
  (fun () ->
     GitHub_mutations.post_comment ~bot_info 
~message:(f "Could not parse bench arguments: %s" err)
~id:comment_info.issue.id
>>= Utils.report_on_postting_comment)
|> Lwt.async;
Server.respond_string ~status:`Ok ~body:"Invalid bench arguments." ()
```

src/webhooks/github.ml
```diff
@@ -93,125 +93,158 @@ let handle_comment_created ~bot_info ~key ~app_id ~github_bot_name
...
+                comment_info )
+      )
+      |> Lwt.async ;
+      Server.respond_string ~status:`OK
+        ~body:(f "Received a request to start the bench.")
+        ()
+    | _ ->
+      Server.respond_string ~status:`OK
+        ~body:(f "Unhandled comment: %s" body)
+        ()
```
lykimq, line 247: Wrong repo/missing install_id/ unknown comment/bad args all hit the same `| _ -> ...` path. 

We could separate "recognized but not allowed" from unknown comments.

```
| Some (`RunCI _ | `Merge | `Bench _) ->
   Server.respond_string ~status:`Ok ~body:"Command recognized but not allowed in this context." ()
| None ->
   Server.respond_string ~status:`Ok ~body:(f "Unhandled comment: %s" body) ()
```

src/webhooks/github.ml
```diff
@@ -93,125 +93,158 @@ let handle_comment_created ~bot_info ~key ~app_id ~github_bot_name
...
+        | args ->
+          match parse_key_value_arguments args with
+          | Result.Ok args -> Some (`Bench args)
+          | _ -> None
+      else
+        None
+    in
+    let parse () =
+      let open Option in
+      let rec first_success body fns =
```
lykimq, line 143: Nit: `first_success` looks equivalent to `List.find_map parsers ~f:(fun parse -> parse body)`, which is already used elsewhere in the repo.

```
let parse () =
  List.find_map 
   [ (fun body ->
         resume_ci_minimize_text_of_body body 
         |> Option.map ~f:(fun x -> `ResumeMinimize x));
      (fun body -> 
         ci_minimize_text_of_body body
         |> Option.map ~f:(fun x -> `Minimize x));
    parse_run_ci;
    parse_merge;
    parse_bench
   ]
 ~f:(fun parse -> parse body)
```

tests/test_string_utils.ml
```diff
@@ -0,0 +1,125 @@
...
+          , check_arguments ~input:"x=foo missing"
+              ~expected:[("x", Some "foo"); ("missing", None)] )
+        ; ( "empty key"
+          , `Quick
+          , check_argument_error ~input:"=value"
+              ~expected:"argument \"=value\" has an empty key" )
+        ; ( "empty quoted key"
+          , `Quick
+          , check_argument_error ~input:{|""|}
+              ~expected:"argument \"\" has an empty key" ) ] ) ]
```
lykimq, line 125: We could add one more bench example that the PR suggested

```
check_argument_error
   ~input:{|coq_opam_packages="a b c" coq_native|}
    ~expected:[("coq_opam_packages", Some "a b c"); ("coq_native", None)]
```

bot-components/utils/String_utils.ml
```diff
@@ -48,6 +48,110 @@ let first_line_of_string s =
...
+              Buffer.add_char buffer escaped ;
+              split (index + 2) quote true tokens )
+            else (
+              Buffer.add_char buffer char ;
+              split (index + 1) quote true tokens )
+      | Some '"' ->
+          Buffer.add_char buffer char ;
+          split (index + 1) quote true tokens
+      | Some _ ->
+          failwith "unsupported quote delimiter"
```
lykimq, line 122: Nit: this arm should be unreachable 
 (`quote` is only `'\''` or `'"'`). Prefer `assert false` over `failwith`, or type quotes as `Single | Double` so the catch-all is unnecessary.

```
| Some _ -> assert false (* quote is only '\'' or '"' *)
```

src/webhooks/github.ml
```diff
@@ -93,125 +93,158 @@ let handle_comment_created ~bot_info ~key ~app_id ~github_bot_name
...
-            |> Lwt.async ;
-            Server.respond_string ~status:`OK
-              ~body:(f "Received a request to start the bench.")
-              () )
-          else
-            Server.respond_string ~status:`OK
-              ~body:(f "Unhandled comment: %s" body)
-              () ) )
+            && Option.is_some install_id ->
+      let key_value_pairs = List.map ~f:(fun (k,v) -> (k, Option.value ~default:"yes" v)) args in
```
lykimq, line 231: Suggestion: alias old `native` to `coq_native` when building job vars.

```
let key_value_pairs =
   List.map args ~f:(fun (k, v) ->
      let k = if String.equal k "native"
                 then "coq_native"
                 else k
       in 
     (k, Option.value ~default:"yes" v)) args
```

src/webhooks/github.ml
Janno: I generalized this slightly so that it can also handle parse errros in `run ci`

bot-components/utils/String_utils.ml
Janno: I used your `Single | Double` suggestion.

src/webhooks/github.ml
Janno: I am not sure this is worth it. The argument is used only rarely and I would advocate for consistency between local benchmark runs and benchmark run triggered through the bot.

Janno:
@lykimq Thank you very much for the detailed review! I think I addressed all comments with code changes except the last one about the alias for `coq_native`. (I haven't marked them resolved in case you want to take another look at the results first.)



> If you want, I can also do a test deployment of this branch so that you can test it in production before we merge.

@Zimmi48  This might be the right way to go. Let's see if anyone has any more feedback first.

Zimmi48:
> >If you want, I can also do a test deployment of this branch so that you can test it in production before we merge.
>
>@Zimmi48 This might be the right way to go. Let's see if anyone has any more feedback first.

Ping me again when it's time to do it. I'm on vacation so I don't look at my emails but I usually receive push notifications for direct mentions.

tests/test_string_utils.ml
```diff
@@ -16,3 +18,132 @@ let () =
   run "String_utils tests"
     [ ( "strip_quoted_bot_name"
       , [test_case "quoted bot name" `Quick test_strip_quoted_bot_name] ) ]
```
lykimq: You have 2 `Alcotest.run`. It will only run this one and the other is never reached.

We should merge both suites into a single `Alcotest.run` call.

src/webhooks/github.ml
```diff
@@ -91,124 +105,152 @@ let handle_comment_created ~bot_info ~key ~app_id ~github_bot_name
...
+        then Some Merge
+        else None
+      in
+      let parse_bench body =
+        if
+          string_match
+            ~regexp:
+              ( f {|@%s:? [Bb]ench\( *$\| +\([^\n]+\) *\(\n\|$\)\)|}
+              @@ Str.quote github_bot_name )
+            body
```
lykimq: The bench regexp uses `[^\n]+`, so only the first line after `bench` is captured. Extra lines are ignored with no error.

Ex:
```
@coqbot bench coq_native
coq_opam_package= "a b c"
```
Only `coq_native` is passed; `coq_opam_packages` is dropped.

We can change by captures everything after `bench` by using `(.+)` instead.

```
( f {|@%s:? [Bb]ench\( *$\| +(.+)|} @@ Str.quote github_bot_name )
```

tests/test_string_utils.ml
Janno: Oh, I see. I had no idea. I merged them.

src/webhooks/github.ml
Janno: My goal was actually to parse arguments only up to an empty line. This way people writing one line between the command and their other text still get a bench run. The text below should work, IMO.

``` 
@coqbot bench
 
Some other text here
```

I have changed the code accordingly and added test cases.

src/utils/bench.ml
```diff
@@ -7,6 +7,77 @@ open HTTP_utils
...
+    | exception _ ->
+        Some (Result.Ok [])
+    | args ->
+        Some
+          (Result.map_error (parse_key_value_arguments args) ~f:(fun error ->
+               f "bench command could not parse key-value arguments: %s" error )
+          )
+  else None
+
+let%test "parses a command without arguments" =
```
lykimq: To keep `bench.ml` focused on its purpose, I would prefer move these inline tests to `tests/test_bench.ml`, as we already do for the parser tests in `tests/test_string_utils.ml`.

Janno: Done!

lykimq:
Thanks for your work. It is good for me now. :)

lykimq (approved):

Janno:
Awesome. Thank you so much for all the feedback!

> Ping me again when it's time to do it. I'm on vacation so I don't look at my emails but I usually receive push notifications for direct mentions.

@Zimmi48 I am ready for a test deployment but it's not at all urgent and definitely not worth interrupting your vacation for. Let's pick this back up when you are back?

Zimmi48:
OK, I'll let you know when I do it.

Zimmi48:
@Janno I am back from my vacation. I can test-deploy whenever is convenient for you to test. Let me know.

PR #376, opened 2026-08-03
Backport and project card logic

lykimq:
- Rename `rocq_push_action` to `push_action` and drive owner/repo/project/org from `repo_config`
- Parameterize `add_to_column` with organization and project number
- Skip backport on push when `github_project_number` is absent; match `PullRequestCardEdited` via install id + project number from config
- Add tests for backport helpers with and without `github_project_number`

Remaining Rocq hardcodes for follow-up PRs.

### DIFF
--- a/src/actions/backport.ml
+++ b/src/actions/backport.ml
@@ -96,3 +96,43 @@ let push_action ~bot_info ~(repo_config : Repo_config.t) ~base_ref ~commits_msg
         else Lwt.return_unit
       in
       Lwt_list.iter_s commit_action commits_msg
+
+let project_action ~bot_info ~(repo_config : Repo_config.t) ~pr_id ~backport_to
+    () =
+  let owner = repo_config.github_owner in
+  let repo = repo_config.github_repo in
+  GitHub_queries.get_pull_request_milestone ~bot_info ~pr_id
+  >>= function
+  | Error err ->
+      Lwt_io.printf "Error: %s\n" err
+  | Ok backport_info -> (
+    match
+      List.find_map backport_info
+        ~f:(fun {backport_to= backport_to'; rejected_milestone} ->
+          if String.equal backport_to backport_to' then Some rejected_milestone
+          else None )
+    with
+    | None ->
+        Lwt_io.printf
+          "PR already not in milestone with backporting info for branch %s.\n"
+          backport_to
+    | Some rejected_milestone -> (
+        Lwt_io.printf
+          "PR is in milestone for which backporting to %s was rejected.\n\
+           Change of milestone requested.\n"
+          backport_to
+        >>= fun () ->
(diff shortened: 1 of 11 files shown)


src/actions/backport.ml
```diff
@@ -4,77 +4,95 @@ open Bot_components.GitHub_types
...
+                      "PR was merged into the backporting branch directly.\n"
+                    >>= fun () ->
+                    GitHub_automation.add_to_column ~bot_info ~organization
+                      ~project:project_number ~backport_to (`PR_ID pr_id)
+                      "Shipped"
+                  else if String.equal base_ref "refs/heads/master" then
+                    (* For now, we hard code that PRs are only backported
                       from master.  In the future, we could make this
                       configurable in the milestone description or in
                       some configuration file. *)
```
Zimmi48, line 43: I guess this is going to be the opportunity to move this to the configuration file (can be in a followup PR).

src/webhooks/github.ml
Zimmi48: Since this call starts to be generic, I think the `project_action` function also needs to be made generic (currently, it still hardcodes Rocq in one place).

By the way, I believe that `project_action` was moved to `bot-components/github/GitHub_automation.ml` by mistake (in a previous PR) and it should probably appear in `src/actions/` instead.

lykimq: I have moved the function `project_action` to `src/action` and fix the missing hardcode Rocq in this commit: [be7b2af](https://github.com/rocq-prover/bot/pull/376/commits/be7b2afe250077cdcb16ac47e2127a547535f3c9)

src/webhooks/github.ml
```diff
@@ -301,23 +301,30 @@ let handle_github_webhook ~bot_info ~key ~app_id ~github_bot_name
...
-             backport_to )
-        ()
+    when String.is_suffix ~suffix:" status" field -> (
+    match
+      Repo_config.find_by_backport_project ~install_id ~project_number
+        repo_config_table
+    with
+    | None ->
+        Server.respond_string ~status:`OK
+          ~body:"Unsupported pull request card edition." ()
```
Zimmi48: I suggest making the message a little different from the standard case, because here it *looks* like a backporting project, but we just don't have the project configured.

src/webhooks/github.ml
```diff
@@ -301,23 +301,30 @@ let handle_github_webhook ~bot_info ~key ~app_id ~github_bot_name
...
+    | Some repo_config ->
+        let backport_to = String.drop_suffix field 7 in
```
Zimmi48, line 317: In passing: we drop a 7-character suffix because we expect the field to be called `<branchname> status`, which we checked above. This might be worth detailing in a comment here, and (later) in documentation.

Zimmi48 (approved):
Nits to (optionally) address before merging.

